PR #2, opened 2021-06-17
Initial ADRs

psmulovics:
Adding
 * 001 Use ADRs
 * 002 Configuration


Sly1024 (commented):
"this would be stored in an in memory data store that could be changed and get notified on the fly"
I'm not sure I agree with that. If the config can be changed, people will use it as a state. State should be separate and persistable. For initial values of some state, it's OK to come from a config field, but I would not allow the config to change.

bingenito:
Think "redux store".  Subsets that make sense to be stored can be persisted back on demand.

psmulovics:
Indeed, I'm with @bingenito on this one. The config can be persisted on demand or even automatically to some medium and read back. This would enable restarting, cross machine manipulation and more.

Sly1024:
That is not config then, it is state. The terminology is confusing. How do you tell which part is persistable/modifiable? I assume developers will want some parts to be fixed, like the title of the main window, the URL for a module, etc.. If we say we have a static config and a persistable state, and we merge them when the app starts, then it's clear what happens and which part is changeable.

bingenito:
Config can and is changed at runtime (by the user changing defaults).  The permissible actions on that are handled by that state providers implementation.  The examples in the ADR are each providers and the traditional configuration in which you are referring is one of them.  Perhaps what needs to be detailed more is an overview of the state being populated by providers and the hierarchical overriding ability.  I would suspect there will be a separate ADR for the individual providers themselves?  The config provider knows how to read and persist config.  Configuration itself mirrors the same concept of hierarchy where you have system readonly, infinite level of readonly config between, and then rw user level.

To an outside observer this is all treated as a single store with events and binding.  You can get away with some interesting fun experiments where you can hibernate the entire state and restore it on restart or keep a side-by-side bcp instance on another pc kept fully in sync via state change broadcasts between each.

bingenito:
@psmulovics  Can you split the first (and less discussed) ADR into a new PR so we can review that separately and individually?

Sly1024:
I had a chat with @psmulovics, I might have missed a discussion about this.
I'm OK with the state being a hierarchical store with binding, notification, etc. But still the terminology is confusing. Is it the "state" or is it the "config" or are these the same thing?

bingenito:
The ADR doesn't have the term "state" in it.  Config would be one portion/provider of this central store.

Sly1024 (approved):
All right, looks good then.

PR #3, opened 2021-06-17
ADR 003 Project Structure

psmulovics:


BalassaMarton:
My input: I'd like to avoid git submodules at all costs. Some git clients can't handle them properly and it's confusing when it comes to stuff like directory.build.props.
Independent pieces of code should be in separate repos so that contributing to one feature/platform/module does not involve cloning the whole universe.

psmulovics:
Agree on avoiding submodules. I'm planning to write about Directory.Build.props/targets and Dependencies.props for the .NET piece in this document mostly.

psmulovics:
@Sly1024 @kruplm let's agree what to say about the frontend part

architecture/adr-003-solution-and-project-structure.md
```diff
@@ -0,0 +1,29 @@
...
+repeatedly entered into the csproj files
+- We would use automated tooling to achieve the maintenance of Dependencies.props file
+
+## Status
+
+Proposed
+
+## Consequences
+
+- We would need to implement an atuoamted tooling for csproj maintainance
```
bingenito, line 27: typo on "atuoamted"

psmulovics: Will add code-spell-checker to extensions.yaml :D

Sly1024:
I agree with having separate repos for each component.
For the front-end, I would not constrain the project structure just yet. It will depend on the best practices for the chosen UI framework (Angular, React, etc.).
However, I think using the Node/NPM ecosystem for building the UI and dependency management is definitely advantageous, and it's the de facto standard in most open source web apps. But I'm not sure if this should be in an ADR about project structure.

BalassaMarton (dismissed):

Sly1024:
As we discussed I think we need ADRs for the frontend (web/UI) parts, about language, frameworks, build infrastructure before we can decide on the project structure. This looks good for the .net projects though.

psmulovics:
Than let's do a quick review on it and I'll merge

Sly1024 (dismissed):
LGTM.

.vscode/extensions.json
```diff
@@ -2,5 +2,9 @@
 
 {
     "recommendations": [
+        "streetsidesoftware.code-spell-checker",
+        "donjayamanne.githistory",
```
bingenito, line 6: Some of these are extremely opinionated, which is fine, but what was the thought behind some?

BalassaMarton: Why are we committing this at all?

psmulovics: So the way how it works, that if you open the project for the first time after cloning, it will popup a "suggested extensions" dialog where you can (if you want) install the extensions from the list. I can remove it if you think this would cause confusion. But yes, these are always going to be opinioned. Same as the upcoming discussion about .editorconfig ;)

psmulovics: So removed this file for now :)

psmulovics:
Can someone from @morganstanley/compose-developers provide a final review before merge?

Sly1024 (approved):

PR #4, opened 2021-06-17
ADR 004 - Module loading

psmulovics:


Sly1024:
How would a module specify its dependencies? In a **manifest file** that we can read before loading the module, or **at runtime**, by calling an API and trying to load the dependency explicitly? 

I can see with the manifest file approach if we have module discoveries registered then we can pre-process the dependency graph before loading any modules. But I think we want to allow loading modules at runtime by an API, which means we cannot discover the dependencies before loading the modules.

How are module discoveries registered? Is that static (coming from config) or can a module register discoveries when it's running? 

How would the module loading process look like (the order of):
- finding/registering module discoveries
- mapping dependencies
- loading/initializing modules

raksanyi:
I feel like we should mention the case when the loaded modules are not the same version. Are we going to provide some kind of backwards-compatibility, or just say it'll only work with a specific version of every module?

Sly1024:
> I feel like we should mention the case when the loaded modules are not the same version. Are we going to provide some kind of backwards-compatibility, or just say it'll only work with a specific version of every module?

If we're going to have separate processes, then the config/API can actually say that it wants to load a module into a specific process. So different versions can be loaded into different processes. If there is a version conflict within the same process, we'll have to resolve that somehow. This is a valid question: do we need the decision about how we resolve version conflicts in this ADR?

psmulovics:
With our plans and previous successes around process isolation, this would not be a problem, @Sly1024 . Regards @raksanyi 's point, we have to provide probably an absolute level of backward compatibility (probably using the same technique as the full .net framework did, e.g. keeping the actual source code of the previous version intact, and version everything around if we to introduce a breaking change).

Sly1024 (approved):
The decisions look solid. I assume we're going to add more ADRs about the specific of module loading and discovery.

PR #5, opened 2021-06-18
Forget to update status on ADRs

psmulovics:
Over time we would get the gist of it, I'm pretty sure :)


bingenito:
@psmulovics This needs to be rebased as we shouldn't be seeing all those other commits imho

bingenito:
I suspect the display of the commits is just reflecting your doing a pull from main into your existing old merged branch which made it look confusing.  Perhaps you should have went with a fresh branch off an up to date main to avoid seeing them.

bingenito:
@psmulovics Should we have a defined time period to keep proposed?

psmulovics:
> @psmulovics Should we have a defined time period to keep proposed?

I don't think so - before I merged I should have made this update. Eg it is in proposed state while it is a PR and when merged it is approved.

bingenito:
For waiting for more feedback. If one reviewer immediately approves and we update and merge it doesn't allow for any comments. I'm wondering if we have a minimum rfc draft period.

psmulovics:
Hm, that's good point. Should we say a week?

psmulovics:
🤕 I'm losing the battle with git

PR #7, opened 2021-06-23
ADR 005 - Messaging

raksanyi:


psmulovics:
LGTM

psmulovics (dismissed):

Sly1024:
Do we need ADR-004 before we merge this?

architecture/adr-005-messaging.md
```diff
@@ -0,0 +1,50 @@
...
+while Cross-machine is between one or more computers.
+A few concrete examples include (but not limited to):
+
+- Notifications
+- Logging, telemetry
+- Querying
+- Pub-Sub
+- Window management
+
+The underlying library that is going to be used is SignalR Core.
```
bingenito, line 26: Should we split this out and explicitly define that it is pluggable and the default out of the box implementation will be SignalR?  It shouldn't be an absolute requirement.

Sly1024: I'm not sure if we need to split it out (do you mean a separate ADR? what would this one be about then?), but I agree that we need to emphasize that the default implementation will use SignalR, but it could be replaced.

bingenito: Explicitly stating as part of the design that it is interface based with an out of box implementation based on SignalR being provided as one such implementation.

bingenito: We shouldn't tie us to infrastructure flavor of the month but rather make it flexible for growth and change.

bingenito: This is worded as though in process communication will still be offloaded via SignalR.  Wouldn't the out of proc communication be an optionally added adaptor to the bus?  You can add any adapter you want and a signalr cross process adapter is one we provide.

psmulovics: This is good. @raksanyi , can you please incorporate this feedback into the ADR?

psmulovics:
> Do we need ADR-004 before we merge this?

Yes, we do

architecture/adr-005-messaging.md
raksanyi: @bingenito Are you suggesting to split this ADR into two, or just change to wording so it's just based on SignalR and we provide one implementation?

psmulovics: I think he meant the latter - e.g it is pluggable and the default implementation would be based on signalr. The pluggable will result in a consequence that the API do not contain any dependency on signalr (why would it have, but better to say explicitely)

architecture/adr-005-messaging.md
```diff
@@ -0,0 +1,54 @@
+# Architecture Decision Record: Messaging
+
+## Context
+
+As [ADR-004](adr-004-module-loading.md) states, ComposeUI needs to be as modular as possible.
+This means any module and functionality should be configurable ([ADR-002](adr-002-configuration.md)).
+One of the core modules of ComposeUI is the messaging module, or the message router. The message router
+should be responsible for every communication that happens between processes.
+It should be an independent pluggable module that runs under the main ComposeUI process, delivering
+messages between processes and computers.
```
bingenito, line 10: Should this read:  and optionally other devices?

architecture/adr-005-messaging.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Decision
+
+Since there are going to be
+multiple kinds of consumers of ComposeUI (Web, WPF, WinForms, ASP.NET, etc) we differentiate two kinds of
```
bingenito, line 15: The term "multiple kinds of consumers" seems off but I don't have an alternative suggestion.

architecture/adr-005-messaging.md
```diff
@@ -0,0 +1,54 @@
...
+messaging: cross-process and cross-machine. Cross-process messaging is strictly between processes on the same PC,
+while Cross-machine is between one or more computers.
```
bingenito, line 17: devices in place of computers?

architecture/adr-005-messaging.md
raksanyi: Actually yes, I guess with web-first there could be other devices.

architecture/adr-005-messaging.md
raksanyi: Maybe "multiple types" or "different types", "different kinds"?

Sly1024: I vote on "different types".

psmulovics (approved):

PR #11, opened 2021-06-29
Adding initial empty solution file with basic Directory.Build content

psmulovics:
The Directory.Build files contain the basic settings for Product name,
authors, company, copyright, email.

Also they contain the setting for debug types (pdb only for release and full for debug).

Also, they remove the target framework from the path,
they generate documentation,
enable pdb files in the package build output folder.

To avoid issues with WPF, it does initialize the LanguageTargets,
and readds xaml and xaml.cs files with the right generator.

Lastly, it contains an empty Dependencies.props file to be filled up by the upcoming tooling.


BalassaMarton (approved):
I'd even add stuff like `TargetFramework`, `Nullable` and other common props here, but looks good to me the way it is now, we can decide on those later.

psmulovics:
Nullable - good idea, TargetFramework - originally I had it there but removed as I could not decide what to put there - 5? already 6?

PR #15, opened 2021-09-08
ADR 008 - Module Security

Sly1024:


psmulovics (approved):

bingenito:
Can we provide at the routing level (as part of the envelope) information about where it came from?  The routing should be able to stamp that in a reliable way?  We could then at the target only accept messages from a source we know.

psmulovics:
> Can we provide at the routing level (as part of the envelope) information about where it came from? The routing should be able to stamp that in a reliable way? We could then at the target only accept messages from a source we know.

So we did had a longer discussion for this. As both sides are local, there is no reliable way to do this :(

bingenito:
> So we did had a longer discussion for this. As both sides are local, there is no reliable way to do this :(

We know the caller at all times whether it is web of .NET.  We need to explore this in more detail as it is a security requirement on other systems to identify callers.

psmulovics:
So we discussed signatures, salting the message, hashes, etc - as long as you are NOT doing .NET 'link demand' remoting, there is no way to enforce that; but we want to avoid doing that under all circumstances.

bingenito:
Let's keep this open as it is a literal showstopper.  It's an internal security requirement that can't be glossed over.  We'll need some type of solution to verify sender.

BalassaMarton:
Can we be more specific about what attacks we are imagining? Eg. user tampering with the config file to load a malicious module, etc. We can then address them one by one.

psmulovics:
I'm all ear how you can prevent XSS if both server and client are local - eg let it be cookies, CSRF, hash, etc everything can be very easily spoofed.

BalassaMarton:
That's still not specific. Who is injecting the code for the XSS attack and where? We should at least ROAM all these risks, and If an attack type is impossible to prevent on the framework level, we should clearly state that.

bingenito:
If module A sends a message to module B to execute a trade, the identify of that sender needs to be verified and approved.

psmulovics:
That would mean we would need to go back to .NET binary formats - and we agreed on we would not do that. Also that would make use of any web client unfeasible.

Sly1024:
Let me talk to a local security guy. I'm interested as well, if there is any way to securely authenticate a client.

psmulovics (approved):

PR #16, opened 2021-10-04
Alternative dockmanager initial PR

psmulovics:
* This would be the alternative dockmanager to the commercial option
* This is an initial PR for it, we would need to figure out the actual folder structure later on
* No tests added as part of this yet


bingenito:
Missing license headers.  Why commit when there is no usage or scenario for using it as an alternative dock manager given there is no shell yet?

psmulovics:
added license headers

src/DockManagerCore/SplitPanes.cs
```diff
@@ -0,0 +1,196 @@
...
+        private SplitPanes first;
+        public SplitPanes First
+        {
+            get => first;
+            set
+            {
+                if (value != first)
+                {
+                    first = value;
+                    if (value != null) value.Parent = null;
```
ztanczos, line 34: Why do you need this line? You're overwriting it right below.

src/DockManagerCore/Utilities/DockingUtils.cs
```diff
@@ -0,0 +1,267 @@
...
+        private static readonly DockingPlaceholder HorizontalDock = new DockingPlaceholder { IsGlobalDocker = true };
+
+        public static FloatingWindow FindWindow(string wid)
+        {
+            WindowCollection windows = Application.Current.Windows;
+            foreach (Window w in windows)
+            {
+                if (w is FloatingWindow && w.Uid.Equals(wid))
+                {
+                    return w as FloatingWindow;
```
ztanczos, line 36: This code can be simplified:
`if (w is FloatingWindow floatingWindow && ...
   return floatingWindow;`

src/DockManagerCore/Desktop/InitialWindowParameters.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// Gets or sets whether the window's header is initially visible.
+        /// </summary>
+        public bool IsHeaderVisible { get; set; }
+
+        public bool Transient { get; set; }
+        /// <summary>
+        /// Gets or sets the resources applied to the window.
+        /// </summary>
+        public ResourceDictionary Resources { get; set; }
```
ztanczos, line 67: Unused property

src/DockManagerCore/Converters/DockLocationConverter2.cs
```diff
@@ -0,0 +1,55 @@
...
+ * or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+using System;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace DockManagerCore.Converters
+{
+    public class DockLocationConverter2:IValueConverter
```
BalassaMarton, line 20: Why the name?

src/DockManagerCore/WindowButton..cs
```diff
@@ -0,0 +1,120 @@
+﻿/*
```
ztanczos, line 1: Nitpick: double '.' in filename: WindowButton..cs

src/DockManagerCore/IPaneContainer.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+
+namespace DockManagerCore
+{
+    public interface IPaneContainer
+    {
+        ContentPane ActivePane { get; } 
+        void AddPane(ContentPane pane_); 
+        IList<ContentPane> GetChildrenPanes(bool leafOnly_=false);
+        IList<PaneContainer> GetChildrenContainers();
```
BalassaMarton, line 23: Rename to `GetChildContainers`?

src/DockManagerCore/SplitPanes.cs
BalassaMarton: It's a little suspicious, isn't it replacing a child element? Should it be this instead?
```cs
                    if (first != null) first.Parent = null;
                    first = value;                 
                    if (first != null) first.Parent = this;
```

src/DockManagerCore/Desktop/PaneFactory.cs
```diff
@@ -0,0 +1,218 @@
...
+    {
+        internal static event RoutedPropertyChangedEventHandler<UIElement> ActivePaneChanged;
+        protected static void HandleActivePaneChanged<TContentPane>(object sender_, TContentPane oldValue_, TContentPane newValue_) where TContentPane : FrameworkElement
+        {
+            if (oldValue_ != null && oldValue_.DataContext is WindowViewModel)
+            {
+                if (LinkControlFocusHackEnabled &&
+                    (oldValue_.DataContext as WindowViewModel).HeaderItems.FirstOrDefault(o => (o is Image) && (o as Image).Tag.ToString() == "LinkControl") != null)
+                {
+                    #region LinkControlFocusHack
```
ztanczos, line 37: hmm :)

PR #30, opened 2022-02-21
Process Explorer API

lilla28:
A basic API for (local) Process Explorer


process explorer/backend/LocalCollector/Connections/IConnectionInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+        public string? LocalEndpoint { get;  set; }
+        public string? RemoteEndpoint { get;  set; }
+        public string? RemoteApplication { get;  set; }
+        public string? RemoteHostname { get;  set; }
+        public ConcurrentDictionary<string, string>? ConnectionInformation { get; set; }
+        public ConnectionStatus GetConnectionStatus(object clientStatus);
+    }
+    public class DummyConnectionInfo : IConnection
+    {
+        public DummyConnectionInfo(ref ClientWebSocket client, Uri remote)
```
ztanczos, line 29: Same comment here: why do you need the ref keyword?

BalassaMarton:
Files are all over the place. Why are a bunch of files under `src/LocalCollector` and others under `process explorer/backend`?

process explorer/backend/LocalCollector/InfoAggregator.cs
```diff
@@ -0,0 +1,59 @@
...
+        public Guid? Id { get; set; } = default;
+        public AppUserInfo User { get; set; }
+        public RegistrationMonitor? Registrations { get; set; } = default;
+        public EnvironmentMonitor? EnvironmentVariables { get; set; } = default;
+        public ConnectionMonitor? Connections { get; set; } = default;
+        public ModuleMonitor? Modules { get; set; } = default;
+        public ProcessMonitor Processses { get; set; }
+
+        //SAMPLE MESSAGE SENDING?
+        public async void SendMessage(string url)
```
ztanczos, line 42: If possible try to avoid async void methods: https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming

process explorer/backend/LocalCollector/Processes/ProcessInfo.cs
```diff
@@ -0,0 +1,227 @@
...
+                    PagedMemoryUsage = process.PagedMemorySize64;
+                    NonPagedMemoryUsage = process.NonpagedSystemMemorySize64;
+                    PagedSystemMemoryUsage = process.PagedSystemMemorySize64;
+                    NonPagedSystemMemoryUsage = process.NonpagedSystemMemorySize64;
+                    PrivateMemoryUsage = process.PrivateMemorySize64;
+                    IsLinux = (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux));
+                    
+                    if (!IsLinux)
+                    {
+                        Children = GetChildProcessesWindows(process);
```
ztanczos, line 58: This class is doing too much. You could use one abstraction to get certain attributes, and that one abstraction could have multiple - OS-specific - implementations

src/LocalCollector/Connections/IConnectionInfo.cs
```diff
@@ -0,0 +1,85 @@
...
+            } }
+        public string? Name { get; set; }
+        public string? LocalEndpoint { get; set; }
+        public string? RemoteEndpoint { get; set; }
+        public string? RemoteApplication { get; set; }
+        public string? RemoteHostname { get; set; }
+        public ConcurrentDictionary<string, string>? ConnectionInformation { get; set; }
+        public Guid? Id { get; init;  }
+
+        private bool SetField<T>(ref T status, T value)
```
ztanczos, line 65: Why do you need this method?

process explorer/backend/LocalCollector/Processes/ProcessInfo.cs
```diff
@@ -0,0 +1,227 @@
...
+                {
+                    ProcessId = process.Id;
+                    ProcessName = process.ProcessName;
+                    PriorityLevel = process.BasePriority;
+                    PagedMemoryUsage = process.PagedMemorySize64;
+                    NonPagedMemoryUsage = process.NonpagedSystemMemorySize64;
+                    PagedSystemMemoryUsage = process.PagedSystemMemorySize64;
+                    NonPagedSystemMemoryUsage = process.NonpagedSystemMemorySize64;
+                    PrivateMemoryUsage = process.PrivateMemorySize64;
+                    IsLinux = (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux));
```
ztanczos, line 54: Why do we care about anything other than Windows? Did I miss something? :)

kruplm:
> Files are all over the place. Why are a bunch of files under `src/LocalCollector` and others under `process explorer/backend`?

She had to create a LocalCollector, which's a separate service. So maybe it's worth to restructure the project a bit. Does it make sense to move the content of '`process explorer/backend/ProcessExplorer` one level up?

process explorer/backend/LocalCollector/Connections/CurrentConnections.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace ProcessExplorer.Entities.Connections
+{
+    public class ConnectionMonitor 
+    {
+        public List<IConnection> connections;
+        public ConnectionMonitor()
+        {
+            connections = new List<IConnection>();
+        }
+        public void AddConnection(ref IConnection connectionInfo)
```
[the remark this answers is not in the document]
lilla28: For one of my following steps, i want to check if the status has been changed and trigger an event or something like that, but i've deleted the ref keyword, because that's not the right way.

process explorer/backend/LocalCollector/InfoAggregator.cs
lilla28: Thanks, i've corrected it! :)

src/LocalCollector/Connections/IConnectionInfo.cs
lilla28: i was playing with INotifyPropertyChanged interface, and that's why i left there -for later work.  But i will delete it for the next commit.

kruplm:
LGTM. If there's nothing else which you'd like to be improved from the C# perspective, then we can merge this.

process explorer/backend/ProcessExplorer/Processes/IProcessGenerator.cs
```diff
@@ -0,0 +1,247 @@
...
+
+    public class ProcessInfoLinux : IProcessGenerator
+    {
+        internal ProcessInfoLinux()
+        {
+
+        }
+        private string RunLinuxPSCommand(Process process, string command)
+        {
+            object locker = new object();
```
ztanczos, line 49: If the intent here is that only one thread should be able to access the critical section below then this wouldn't work since each thread would create its own 'locker' instance.

ztanczos:
I'd recommend a 'one class per file' rule (except for private / nested classes).

process explorer/backend/LocalCollector/Connections/Connection.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        public ConnectionDto Data { get; set; }
+
+        public string StatusChanged(WebSocketState clientStatus)
+        {
+            var clientState = clientStatus;
+
+            if (clientState == WebSocketState.Open)
+                return ConnectionStatus.Running.ToString();
```
ztanczos, line 66: enum.ToString is a surprisingly slow operation. I'd recommend a different approach here.

process explorer/backend/LocalCollector/InfoAggregator.cs
```diff
@@ -0,0 +1,69 @@
...
+                if (responseMessage.IsSuccessStatusCode)
+                {
+                    Uri? infoUrl = responseMessage.Headers.Location;
+                    Console.WriteLine(infoUrl);
+                }
+            }
+        }
+        public void Send(string url)
+        {
+            var task = SendMessage(url);
```
ztanczos, line 55: What's the intent here? Here you're essentially making an asynchronous call synchronous by blocking the caller thread.

process explorer/backend/LocalCollector/InfoAggregator.cs
```diff
@@ -0,0 +1,69 @@
...
+            Data.Registrations = registrations;
+            Data.Modules = modules;
+        }
+        
+        public InfoAggregatorDto Data { get; set; }
+
+        //SAMPLE MESSAGE SENDING
+        private async Task SendMessage(string url)
+        {
+            using (var client = new HttpClient())
```
ztanczos, line 39: I'm not sure I understand the context here but it seems you're using HttpClient wrong: see https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests for details

process explorer/backend/LocalCollector/User/Machine.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                Data.MachineName = Environment.MachineName;
+                Data.IsUnix = (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux));
+                Data.OSVersion = Environment.OSVersion;
+                Data.Is64BIOS = Environment.Is64BitOperatingSystem;
+                Data.Is64BitProcess = Environment.Is64BitProcess;
+            }
+            catch (Exception ex)
+            {
+                throw new Exception(ex.Message);
```
ztanczos, line 42: Don't do this. This way you're losing the original exception, the stacktrace, etc. If you can't/don't want to do anything with the exception then don't catch it.

process explorer/backend/ProcessExplorer/Processes/IProcessGenerator.cs
```diff
@@ -0,0 +1,245 @@
...
+                    }
+                }
+            }
+            return list;
+        }
+    }
+
+    public class ProcessInfoLinux : IProcessGenerator
+    {
+        object locker = new object();
```
ztanczos, line 44: private readonly

process explorer/backend/ProcessExplorer/Processes/IProcessGenerator.cs
```diff
@@ -0,0 +1,245 @@
...
+                "Where TargetInstance ISA 'Win32_Process' ";
+
+                Watcher = new ManagementEventWatcher(Scope, new EventQuery(WmiQuery));
+                Watcher.EventArrived += new EventArrivedEventHandler(WmiEventHandler);
+                Watcher.Start();
+                //Watcher.Stop();
+            }
+            catch (Exception)
+            {
+                throw new ApplicationException("Error while reading the processes.");
```
ztanczos, line 216: https://docs.microsoft.com/en-us/dotnet/api/system.applicationexception?view=net-6.0
"You should not throw an [ApplicationException](https://docs.microsoft.com/en-us/dotnet/api/system.applicationexception?view=net-6.0) exception in your code,"

process explorer/backend/LocalCollector/EnvironmentVariables/EnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class EnvironmentMonitorDto
+    {
+        public ConcurrentDictionary<string, string>? EnvironmentVariables { get; set; } = new ConcurrentDictionary<string, string>();
+
+        public static EnvironmentMonitorDto FromEnvironment()
+        {
+            var envs = new EnvironmentMonitorDto();
+            if (envs.EnvironmentVariables != null)
+            {
+                lock (envs.EnvironmentVariables)
```
ztanczos, line 17: Guidelines
When you synchronize thread access to a shared resource, lock on a dedicated object instance (for example, private readonly object balanceLock = new object();) or another instance that is unlikely to be used as a lock object by unrelated parts of the code. 
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock

process explorer/backend/ProcessExplorer/Processes/ProcessMonitor.cs
```diff
@@ -0,0 +1,375 @@
...
+        private async void TryDeleteChildProcesses(int pid)
+        {
+            try
+            {
+                lock (Data.Processes)
+                {
+                    foreach (var process in Data.Processes)
+                    {
+                        if (process.Children != default)
+                            lock (process.Children)
```
Sly1024, line 183: I don't think we need a second lock here. There's a lock on a common object (Data.Processes) which prevents other threads from entering this block anyway.

process explorer/backend/LocalCollector/Connections/ConnectionMonitor.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public ConnectionMonitorDto Data { get; set; } = new ConnectionMonitorDto();
+        public ConnectionMonitor()
+        {
+            Data.Connections = new SynchronizedCollection<ConnectionDto>();
+        }
+        public ConnectionMonitor(SynchronizedCollection<ConnectionDto> connections)
+            => this.Data.Connections = connections;
+        public void AddConnection(ConnectionDto connectionInfo)
+            => Data?.Connections?.Add(connectionInfo);
```
ztanczos, line 17: The null-checks are inconsistent in this class. Either you should check for null everywhere if that's a valid state or if this.Data cannot be null then you should remove the null checks.

process explorer/backend/LocalCollector/Modules/ModuleMonitorDto.cs
```diff
@@ -0,0 +1,30 @@
...
+            var monduleMonitor = new ModuleMonitorDto();
+            var assemblies = AppDomain.CurrentDomain.GetAssemblies();
+            lock (locker)
+            {
+                foreach (var assembly in assemblies)
+                {
+                    var modules = assembly.GetLoadedModules();
+                    foreach (var module in modules)
+                    {
+                        monduleMonitor?.CurrentModules?.Add(ModuleDto.FromModule(assembly, module));
```
ztanczos, line 23: moduleMonitor cannot be null here.

process explorer/backend/ProcessExplorer/Processes/ProcessGeneratorBase.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// </summary>
+        /// <param name="process"></param>
+        /// <returns></returns>
+        public bool IsComposeProcess(object process)
+        {
+            try
+            {
+                var proc = process as Process;
+                if (proc is not null && ProcessMonitor.ComposePID != default)
+                    while (proc?.Id != 0)
```
ztanczos, line 152: You checked that proc is not null one line above.

process explorer/backend/ProcessExplorer/Processes/ProcessMonitor.cs
```diff
@@ -275,15 +291,24 @@ private async Task<bool> TryDeleteMainProcesses(int pid)
...
+                Data.Processes[index].ProcessStatus = Status.Terminated.ToStringCached();
+            }
+        }
+
         /// <summary>
         /// Delays the removing process from the list.
         /// </summary>
         /// <param name="item"></param>
-        private async void DeleteFromListAfter1Minute(ProcessInfoDto item)
+        private void DeleteFromListAfter1Minute(ProcessInfoDto item, int delay)
```
Sly1024, line 353: I would remove the "1Minute" from the name here. Maybe "RemoveAfterTimeout".
It seems that the "DelayTime" is accessible from this method, so you don't need to pass it as an argument.

process explorer/backend/ProcessExplorer/Processes/Interfaces/IProcessMonitor.cs
```diff
@@ -15,6 +15,7 @@ public interface IProcessMonitor
         void KillProcessByName(string processName);
         void SetSubribeUrl(string url);
         void SetComposePID(int pid);
+        void SetDelay(int delay);
```
Sly1024, line 18: I think the word "delay" alone does not communicate what is being delayed. We need something like "SetDeadProcessRemovalDelay".

process explorer/backend/ProcessExplorer/Processes/ProcessMonitor.cs
```diff
@@ -17,6 +17,11 @@ public class ProcessMonitor : IProcessMonitor
         /// </summary>
         public static int ComposePID { get; set; }
 
+        /// <summary>
+        /// Delay time for keeping the process in the list after it is terminated.
+        /// </summary>
+        public static int DelayTime { get; set; } = 60000;
```
Sly1024, line 24: I would add the unit in the comment to make it clear that this number is interpreted as **milliseconds**, not seconds or minutes, etc.
That brings up the other question: Do we want this to be controlled with millisecond precision? I think **seconds** would be a better unit.

PR #42, opened 2022-03-18
Environment variables/dialog window

kruplm:


psmulovics:
I like where this is going, @Sly1024 , any comments from you?

kruplm:
Any other feedback from @Sly1024 or anyone else?
I'd like to merge this EOD if there're no objections.

PR #45, opened 2022-05-23
Implementing the functionality of ProcessInfoCollector

lilla28:
InfoCollector contains information of environment variables, loaded modules, registrations, and connections.

- IProcessInfoCollector: collects the data of environment variables, loaded modules, registrations, and connections. Sends message to the backend if something is changed.
- ProcessInfoCollectorData: contains the runtime information of the process. Also updates/adds information into the collections.
- Sends update and message to the backend PE.
- Collects the relevant information.


process explorer/backend/LocalCollector/Connections/ConnectionMonitor.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        public void AddConnections(SynchronizedCollection<ConnectionInfo> connections)
+        {
+            foreach (var conn in connections)
+            {
+                lock (locker)
+                {
+                    var element = Data.Connections.FirstOrDefault(item => item.Id == conn.Id);
+                    var index = Data.Connections.IndexOf(conn);
```
Sly1024, line 53: it should be `IndexOf(element)` here.

process explorer/backend/LocalCollector/Connections/ConnectionMonitor.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                Data.Connections.Remove(connectionInfo);
+            }
+        }
+
+        public void AddConnections(SynchronizedCollection<ConnectionInfo> connections)
+        {
+            foreach (var conn in connections)
+            {
+                lock (locker)
```
Sly1024, line 50: Do we want to lock for each element in the `connections` list? Maybe we can move the lock outside of the foreach. Not sure if that's better to lock for the entire time while we update the `Connections` list..

process explorer/backend/LocalCollector/Connections/ConnectionMonitor.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                Data.Connections.Add(connectionInfo);
+            }
+        }
+
+        public void RemoveConnection(ConnectionInfo connectionInfo)
+        {
+            lock (locker)
+            {
+                Data.Connections.Remove(connectionInfo);
```
Sly1024, line 42: Should we check for a `ConnectionInfo` with the same ID? This only removes it if the exact same object is in the collection.

process explorer/backend/LocalCollector/ProcessInfoCollector.cs
```diff
@@ -0,0 +1,259 @@
...
+            this.logger = logger;
+
+            if (assemblyId is not null)
+            {
+                this.assemblyID.Name = assemblyId;
+            }
+
+            if (pid is not null)
+            {
+                Data.Id = Convert.ToInt32(pid);
```
Sly1024, line 44: I would think that if the type of `pid` is `int?` and it `is not null` then it must be `int`, so there should be no need to convert.

Sly1024 (changes requested):

process explorer/backend/LocalCollector/ProcessInfoCollector.cs
lilla28: I think we won't need neccessarily to give the pid to the ctor. 
If I use type int?, then I should convert explicitly to int, because it cannot  do that implicitly after I check.


_By default it will read the pid if it is null -in the ProcessInfoCollectorData. 
 _public int Id { get; set; } = Process.GetCurrentProcess().Id;__

process explorer/backend/LocalCollector/Connections/ConnectionMonitor.cs
lilla28: I changed it, thanks for highlighting! :)

process explorer/backend/LocalCollector/ProcessInfoCollector.cs
Sly1024: Sorry, you're right. I forgot that it's a value type, so it gets boxed, that's why it doesn't convert implicitly.

Sly1024 (approved):

psmulovics (approved):

PR #46, opened 2022-06-13
ADR-010 - Messaging Layers

Sly1024:


architecture/adr-010-messaging-layers.md
```diff
@@ -0,0 +1,60 @@
...
+* `void RegisterConverter<TCustom>(IConverter<TCustom> converter)`
+
+Here the converter is for a custom type that is passed as (part of) the payload, 
+but the serializer converts the whole message object that the higher level messaging API sends.
+
+### Channel API
+
+The channel API provides the functionality to send and receive packages asynchronously.
+
+* `Task Connect()`
```
ztanczos, line 49: Minor comment about naming: All these methods are async but the naming is inconsistent: some are postfixed with 'Async' some are not.

architecture/adr-010-messaging-layers.md
```diff
@@ -0,0 +1,60 @@
...
+* Channel
+
+### Messaging API
+
+The Messaging API is a high level, object based communication API with the following features.
+
+* Provides a publish-subscribe communication API
+    * `Task Publish<TMessage>(string topic, TMessage message)`
+    * `Task<IObservable<TMessage>> Subscribe<TMessage>(string topic)`
+    * `UnSubscribe(string topic)`
```
ztanczos, line 26: Nitpick: 'unsubscribe' is one word so the correct casing would be 'Unsubscribe'

architecture/adr-010-messaging-layers.md
bingenito: https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap

architecture/adr-010-messaging-layers.md
```diff
@@ -0,0 +1,60 @@
...
+* Provides a request-response type service calls
+    * `Task RegisterQueryService<TResult>(string serviceId, IServiceHandler<TResult> handler)`
```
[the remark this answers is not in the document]
Sly1024: For one `serviceId` there can be only one "responder" and I've added an Unregister function, but we need to make sure only the module that did the registration can unregister.

architecture/adr-010-messaging-layers.md
```diff
@@ -0,0 +1,61 @@
...
+* Channel
+
+### Messaging API
+
+The Messaging API is a high level, object based communication API with the following features.
+
+* Provides a publish-subscribe communication API
+    * `Task PublishAsync<TMessage>(string topic, TMessage message)`
+    * `Task<IObservable<TMessage>> SubscribeAsync<TMessage>(string topic)`
+    * `Task UnsubscribeAsync(string topic)`
```
BalassaMarton, line 26: This feels unintuitive to me. To subscribe, I'd have to do this:
```cs
var subscription = (await messageRouter.SubscribeAsync(topic)).Subscribe(new Observable(msg => HandleMessage(msg)));
```
`IObservable<T>.Subscribe` returns an `IDisposable` for unsubscribing, so what's the `Unsubscibe(topic)` method doing here?

architecture/adr-010-messaging-layers.md
```diff
@@ -0,0 +1,61 @@
...
+    * `Task RegisterQueryResponderAsync<TResult>(string serviceId, IServiceResponder<TResult> responder)`
+    * `Task UnregisterQueryResponderAsync<TResult>(string serviceId)`
+    * `Task<TResult> QueryAsync(string serviceId, IServiceParams params)`
+* Provides a way to swap out the default serialization and channel implementations
+    * `void SetMessageSerializer(ISerializer serializer)`
+    * `void SetMessageChannel(IChannel channel)`
+
+### Serialization API
+
+The serialization layer provides the functionality to convert message objects to byte sequences and vice versa.
```
BalassaMarton, line 37: Haven't we already rejected the idea of strongly typed messages? How would this work for JS and other clients? Who's responsibility is it to implement the serializer(s) for every language binding?

architecture/adr-010-messaging-layers.md
```diff
@@ -0,0 +1,61 @@
...
+The Messaging API is a high level, object based communication API with the following features.
+
+* Provides a publish-subscribe communication API
+    * `Task PublishAsync<TMessage>(string topic, TMessage message)`
+    * `Task<IObservable<TMessage>> SubscribeAsync<TMessage>(string topic)`
+    * `Task UnsubscribeAsync(string topic)`
+* Provides a request-response type service calls
+    * `Task RegisterQueryResponderAsync<TResult>(string serviceId, IServiceResponder<TResult> responder)`
+    * `Task UnregisterQueryResponderAsync<TResult>(string serviceId)`
+    * `Task<TResult> QueryAsync(string serviceId, IServiceParams params)`
```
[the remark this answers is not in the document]
Sly1024: I don't like the word "invoke" to be honest. It feels more natural that a service should be called or queried. 
Others, do you have any strong opinion?

architecture/adr-010-messaging-layers.md
Sly1024: I think we could use a JSON serializer by default (Newtonsoft, Microsoft), which can automatically convert DTO-like objects with properties.

Sly1024: I think the main point here is that since the message is converted to JSON, we don't require the actual types to match on the two sides. They just need to be compatible (same property names/types).

architecture/adr-010-messaging-layers.md
Sly1024: I've changed it as I suggested, plus I wrapped the `TMessage` in a `RouterMessage` to include metadata.

BalassaMarton: We can still pass an `IObserver` to subscribe, that can receive error and completion events.
```cs
Task<IDisposable> SubscribeAsync<TMessage>(IObserver observer);
```
OR we can still use observables and let users leverage reactive extensions like they used to:
```cs
var subscription = messageRouter.Topic<TMessage>(topic).Subscribe(observer);
```
Note that there's now a `System.Reactive.Async` package which has `IAsyncObservable` & `IAsyncObserver`, with LINQ for `IAsyncEnumerable` and everything you can imagine. So it might make sense to just have an API facede that provides the sync and async API for subscriptions:
```cs
public sealed class Topic<TMessage> : IObservable<TMessage>, IAsyncObservable<TMessage>
{ ... }
```

architecture/adr-010-messaging-layers.md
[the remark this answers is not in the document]
ztanczos: Just an idea: since the idea here is already to provide a request-response type service calls what if we'd call it 'Request' or 'SendRequest'?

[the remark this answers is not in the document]
Sly1024: How would you call the "service"? RequestResponder, Handler, Service?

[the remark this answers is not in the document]
BalassaMarton: Yeah we should agree on terminology and abstractions before defining APIs.

architecture/adr-010-messaging-layers.md
BalassaMarton: Given that messages are plain text (as per ADR-009, property bags/key-value pairs to be more precise), each API will have its own serialization format, if any. It doesn't make sense to me to register a global serialization handler. We can add extensions to the router client that handle (de)serialization of specific formats, but this should not be part of the core API. For reference, see the [HttpClientJsonExtensions](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.json.httpclientjsonextensions?view=net-6.0) class from .NET.
Personally, I'd be more happy with a standard, structured format such as JSON across the whole platform, to avoid this hell that will surely arise when every app developer starts inventing their own format. It's easy to overlook this problem when we look at JS-based frameworks where everything runs on the same platform and they just pass around objects without even considering how those are marshaled to each process.

architecture/adr-010-messaging-layers.md
```diff
@@ -0,0 +1,63 @@
...
+* `ReadOnlyMemory<byte> Serialize(object message)`
+* `object Deserialize(ReadOnlyMemory<byte> message)`
+* `void RegisterConverter<TCustom>(IConverter<TCustom> converter)`
+
+Here the converter is for a custom type that is passed as (part of) the payload, 
+but the serializer converts the whole message object that the higher level messaging API sends.
+
+### Channel API
+
+The channel API provides the functionality to send and receive packages asynchronously.
```
BalassaMarton, line 50: I can't see where routing information is provided. Is it part of the message format?

BalassaMarton: IMHO routing info should be a separate parameter or these methods should receive some structure that encapsulates routing info and message bytes (instead of just a byte sequence).

architecture/adr-010-messaging-layers.md
Sly1024: My idea was that the Message Router will have its own message format/types. For example:
```cs
public record RouterMessage<TMessage>(string id, string sender, MessageType type, 
    MessageParams params, TMessage payload);
```
When you call a "publish" it would send a message like this:
```cs
public Task PublishAsync<TMessage>(string topic, TMessage user_payload) {
    var message = new RouterMessage("123", "compose-client-567", MessageType.Publish, 
        new MessageParams { Topic = "topic"}, user_payload);
    var serializedMessage = serializer.Serialize(message);
    return channel.SendAsync(serializedMessage);
}
```
So it would wrap the `user_payload` with our own message to provide meta data. To serialize that, we need a serializer component, because I don't want to enforce the users to use a specific serializer (Newtonsoft). If we use a serializer, then it can also be used to serialize the custom payload, so why would we not allow the user to pass in an object, as long as it can be serialized by the serializer (the default or a custom one).

In contrast, If we say we only allow strings, then it means our messages can be strings only, and I don't think this is how we should do it:
```cs
channel.Send($"{ id: '{id}', sender: '{sender}', msgType: 'publish', 
    topic: '{topic}', payload: '{payload}' }");
```

architecture/adr-010-messaging-layers.md
Sly1024: I thought of the "channel" as a single connection between two endpoints. So each client will have its own channel (that wraps the websocket) and the Router will know to which channel it needs to send the message. The "routing" is done by the Router.

architecture/adr-010-messaging-layers.md
```diff
@@ -0,0 +1,68 @@
...
+
+* `Task ConnectAsync()`
+* `Task SendAsync(ReadOnlyMemory<byte> package)`
+* `Task<ReadOnlyMemory<byte>> ReceiveAsync()`
+* `Task DisconnectAsync()`
+
+On the Router side, it also needs to act as a "server" to which the clients connect.
```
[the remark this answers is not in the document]
Sly1024: OK, let's say we make two classes a `ChannelServer` and a `ChannelClient`. The Router runs in Component1, and Component2 connects to it. Then the init code in Component1 (can be the main/core component) calls `StartServer()` on the `ChannelServer` which would open a websocket server and listen to the connecting clients. When Component2 wants to connect, it creates a `ChannelClient` and calls `ConnectAsync()` and then the server fires the `ClientConnected` event with the args containing the `ChannelClient` instance that represents the server end of the websocket connection. So there is a `ChannelClient` on both sides that are the two ends of the same connection.

architecture/adr-010-messaging-layers.md
BalassaMarton: But won't this kind of serialization just create a dependency hell in the consuming project, because now it has to know what serializer was configured in the host, and it might not even have a language binding for that particular serializer? If in contrast a module's public API (basically all the query responders) defines its own message format, then there's a contract between a specific version of a module and its clients (all the components that interact with it). This is basically the web, so we might as well just create an in-memory Internet 😄

Sly1024: I was thinking about this from another angle: what if two modules want to communicate using a custom format, and they can agree on the serializer, but the rest of the Compose modules expect JSON? Should we allow customizing the serializer for each module? The wrapper messages must be in a format that the Router understands. In this case the two modules could use their own serializer and just pass in a string or a byte array as payload, then the rest can still be JSON. 

I'm tempted to say we should accept only `string` as payload, or maybe `object` and allow registering serializers for the payload part only.

architecture/adr-010-messaging-layers.md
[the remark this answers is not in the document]
Sly1024: I've added a diagram, but I'm not sure if it tells the story well.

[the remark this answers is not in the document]
Sly1024: Let me explain it with code.

Router:
```cs
var server = new ChannelServer(port: 5678);
server.ClientConnected += (object sender, ClientConnectedArgs args) => {
  ChannelClient client = args.Client;
  if (await AuthenticateClient(client)) {
    connectedClients.Add(client);
  } else {
    await client.DisconnectAsync();
  }
};
server.StartServer();
```

Client Module:
```cs
var client = new ChannelClient("localhost:5678");
await client.ConnectAsync();
await HandleAuth(client);
await client.SendAsync(...);
```

architecture/adr-010-messaging-layers.md
BalassaMarton: Just forget the whole "registering stuff" part. Anything you register in one module, needs to be registered the same way in all the other modules. This is true for payloads or the protocol frame. This would completely break composability, as you can't just write a component in any language and drop it into the application via a few lines of config, you now have to make sure that whatever serializer and channel implementation you have chosen in the host is also available and configured in each of the components. I strongly vote against it.

architecture/adr-010-messaging-layers.md
BalassaMarton: For customizable channels: if you look at SignalR or Dapr, the protocol is usually negotiable, meaning that the server and client can support multiple transport protocols and they negotiate which one to use, automatically (for SignalR it was WebSockets, long polling and now gRPC, for Dapr it's HTTP and gRPC). All this information - connection info for each available endpoint - can be passed to the module at run time, without the need for manual configuration. We don't even have to chose one channel implementation, they can run side by side.

Sly1024:
We decided to go with a PoC first, so I'm closing this PR for now.

PR #49, opened 2022-07-06
Compose code skeleton

nick-polyak-ms:
Created a code skeleton and a number of plugin samples. Please review


bingenito:
License header missing on all source files

BalassaMarton (approved):

NOTICE
```diff
@@ -2,3 +2,11 @@ Morgan Stanley ComposeUI
 Copyright 2021 Morgan Stanley
 
 This product includes software developed at Morgan Stanley (http://www.morganstanley.com).
+
+=======================================================================
+3rd party software used by ComposeUI
```
bingenito, line 7: Did Legal request/review this?

src/MorganStanley.ComposeUI.MainApp/App.axaml.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Avalonia;
```
bingenito, line 1: Source files in this project are missing license headers

nick-polyak-ms: I'll check them again

src/Common/MorganStanley.ComposeUI.Common.Interfaces/ISyncContext.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace MorganStanley.ComposeUI.Common.Interfaces
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public interface ISyncContext
+    {
+        /// <summary>
+        ///
```
bingenito, line 20: Empty summaries

src/MorganStanley.ComposeUI.MainApp/App.axaml.cs
nick-polyak-ms: fixed

src/Common/MorganStanley.ComposeUI.Common.Interfaces/ISyncContext.cs
nick-polyak-ms: fixed

NOTICE
nick-polyak-ms: What are we going to do about it? Should I do anything?

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/DispatcherAdapter.cs
```diff
@@ -0,0 +1,51 @@
...
+using MorganStanley.ComposeUI.Common.Interfaces;
+using NP.Utilities.Attributes;
+
+
+namespace MorganStanley.ComposeUI.Common.VisualUtils
+{
+    /// <summary>
+    /// this class implements Avalonia UI thread marshalling functionality. 
+    /// </summary>
+    [Implements(typeof(ISyncContext), isSingleton:true)]
```
ztanczos, line 22: What's this attribute?

src/Common/MorganStanley.ComposeUI.Common.Utilities/MorganStanley.ComposeUI.Common.Utilities.csproj
```diff
@@ -0,0 +1,25 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NP.Concepts.Behaviors" Version="1.0.9" />
```
ztanczos, line 22: What's this package?

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/MorganStanley.ComposeUI.Common.VisualUtils.csproj
```diff
@@ -0,0 +1,30 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NP.Avalonia.Visuals" Version="1.0.24" />
```
ztanczos, line 22: What does this package do?

ztanczos:
Nitpick: you renamed Directory.Build.targets to Directory.Build.targets_SAVE - was that intentional?

Playground/Plugins/Services/MockAuthentication/MockAuthentication.csproj
```diff
@@ -0,0 +1,31 @@
...
+-->
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <Target Name="PostBuild" AfterTargets="PostBuildEvent">
```
psmulovics, line 21: Can we add here (and each cases we do that) an extra check and breaking the build if something is wrong? e.g. does $(SolutionDir)\bin\$(Configuration)\net6.0\Plugins\Services\ exist?

ZKRobi:
In my opinion whole "Playground" and "Prototypes" folder structure goes against the idea of using a version control system. If we want to focus efforts on a certain topic, we should create a branch or fork for that effort instead of putting it in the main branch. Basically the whole "common" vs "playground" separation with the idea of putting only stable code in the "common" part is, in reality, a second layer of merging and PRs built upon the already existing system. We should use the existing versioning system to solve code management, not a folder structure.

ZKRobi:
I gave some thought to the topic of "shell" and "core" separation, and I highly disagree with the statement this is an abstraction we can make later. This separation is one of the main features of ComposeUI, and thus, it has to be represented from the start in the main branch.
Core code shouldn't depend on any UI framework, that's the job of the replacable shell.
At the moment it's way too soon to choose any DI container technology. I'm not even convinced it's something we'll ever need in the main compose project. Configurations where the same thing can have multiple implementations and compositions etc is just asking for trouble with a DI configuration. Using a DI container should be for convenience, not for the sake of using a DI container.

ZKRobi:
I'm worried about the idea of in-process plugins. It kind of requires a locked UI technology, it opens up place for a whole world of abusing access to in-process objects, so in general it's a much harder topic than it seems on the surface.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/DispatcherAdapter.cs
BalassaMarton: This is one of my main concerns - we shot ourselves in the foot with Unity because it was owned by a single person. I don't think we should go with a DI container nobody knows. Not because I don't trust it but because it's against the design principle of least dependencies. If we still go with this DI solution, we must provide adapters for plugins that want to use the de facto standard MSFT extension.

src/Common/MorganStanley.ComposeUI.Common.Utilities/MorganStanley.ComposeUI.Common.Utilities.csproj
BalassaMarton: +1 on this. Can we include less dependencies on packages maintained by a single person?

BalassaMarton:
> I'm worried about the idea of in-process plugins. It kind of requires a locked UI technology, it opens up place for a whole world of abusing access to in-process objects, so in general it's a much harder topic than it seems on the surface.

I disagree. It does not require any technology locking as we will define the abstractions in a framework-agnostic way. Access to in-process objects is a necessity for plugins. We expect very few externally developed plugins and a lot of modules instead.

BalassaMarton:
> In my opinion whole "Playground" and "Prototypes" folder structure goes against the idea of using a version control system. If we want to focus efforts on a certain topic, we should create a branch or fork for that effort instead of putting it in the main branch. Basically the whole "common" vs "playground" separation with the idea of putting only stable code in the "common" part is, in reality, a second layer of merging and PRs built upon the already existing system. We should use the existing versioning system to solve code management, not a folder structure.

I agree in general, but we are now in a prototyping phase when frequent code merges are normal, and branching out would just make this way harder with all the merge conflicts. Later we can clear this up and remove all the non-working stuff, once we have regular RTM releases.

src/Common/MorganStanley.ComposeUI.Common.BasicModels/MorganStanley.ComposeUI.Common.BasicModels.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<!--
```
BalassaMarton, line 1: This package seems unnecessary, should be merged into a `MorganStanley.ComposeUI.Core` package with all the abstractions and types commonly used by all plugins.

src/Common/MorganStanley.ComposeUI.Common.Interfaces/MorganStanley.ComposeUI.Common.Interfaces.csproj
```diff
@@ -0,0 +1,21 @@
...
+Morgan Stanley makes this available to you under the Apache License, Version 2.0 (the "License").
+You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+See the NOTICE file distributed with this work for additional information regarding copyright ownership.
+Unless required by applicable law or agreed to in writing, software distributed under the License
+is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and limitations under the License.
+
+********************************************************************************************************
+-->
+<Project Sdk="Microsoft.NET.Sdk">
```
BalassaMarton, line 13: This package seems unnecessary, should be merged into a `MorganStanley.ComposeUI.Core` package with all the abstractions and types commonly used by all plugins.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/MorganStanley.ComposeUI.Common.VisualUtils.csproj
```diff
@@ -0,0 +1,30 @@
...
+Morgan Stanley makes this available to you under the Apache License, Version 2.0 (the "License").
+You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+See the NOTICE file distributed with this work for additional information regarding copyright ownership.
+Unless required by applicable law or agreed to in writing, software distributed under the License
+is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and limitations under the License.
+
+********************************************************************************************************
+-->
+<Project Sdk="Microsoft.NET.Sdk">
```
BalassaMarton, line 13: I'd name this project differently as it is Avalonia-specific. Maybe we should group packages by UI technology and put them into respective folders.

BalassaMarton:
Please remove junk files like `src/Directory.Build.targets_SAVE`

BalassaMarton:
The folder structure is inconsistent. Bare in mind that we will have javascript/node projects as well. Under `src` we should have stuff like `core` (contains the .NET solutions providing core functionality and default plugin implementations), `avalonia-shell` (projects for the Avalonia-based shell), `javascript-binding` (npm packages for JavaScript modules), etc.

ZKRobi:
> 

I can live with it either way, I'd prefer using a prototyping branch for this purpose :)

ZKRobi:
> > I'm worried about the idea of in-process plugins. It kind of requires a locked UI technology, it opens up place for a whole world of abusing access to in-process objects, so in general it's a much harder topic than it seems on the surface.
> 
> I disagree. It does not require any technology locking as we will define the abstractions in a framework-agnostic way. Access to in-process objects is a necessity for plugins. We expect very few externally developed plugins and a lot of modules instead.

I'm not against the idea of making this a possibility, but I'd definiately not include it initially. In my opinion there are way too many open questions for this to be a worthwile development angle at the moment, with way too many things changing.
I'd return to the topic of in-process plugins when we have a solid application lifecycle and core + shell architecture in place.

ZKRobi:
> > In my opinion whole "Playground" and "Prototypes" folder structure goes against the idea of using a version control system. If we want to focus efforts on a certain topic, we should create a branch or fork for that effort instead of putting it in the main branch. Basically the whole "common" vs "playground" separation with the idea of putting only stable code in the "common" part is, in reality, a second layer of merging and PRs built upon the already existing system. We should use the existing versioning system to solve code management, not a folder structure.
> 
> I agree in general, but we are now in a prototyping phase when frequent code merges are normal, and branching out would just make this way harder with all the merge conflicts. Later we can clear this up and remove all the non-working stuff, once we have regular RTM releases.

I can live with it either way, in my opinion what you are describing is a "feature branch" (like prototypes/hackday1) and merge conflicts will happen anyway. Both approaches have their merits, I'll live with whichever the team decides :)

ZKRobi:
A slightly off-topic comment, and maybe a bit early, but do we want to have some common requirements on commit/merge contents stlye?
Some examples:
Rebase vs merge commits (Some teams prefer a linear history, some prefer seeing "what really happened")
What should be a commit (In some teams, each commit should be valuable on its own)
How many commits should belong to a single feature (some teams like to merge all the changes in a single commit, others like to display progress through multiple ones)

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/DispatcherAdapter.cs
nick-polyak-ms: I'll be happy to replace it with MEF, but standard MSFT has severe limitations and we should avoid using it.

src/Common/MorganStanley.ComposeUI.Common.Utilities/MorganStanley.ComposeUI.Common.Utilities.csproj
nick-polyak-ms: Avalonia needs some third party tooling, including UniDock. UniDock depends on this package as well as NP.Avalonia.Visuals and NP.Utilities. This is all open source with plenty of samples. There are no other Docking Avalonia packages that do the job, so we are kind of forced to use it.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/MorganStanley.ComposeUI.Common.VisualUtils.csproj
nick-polyak-ms: This package replaces 3rd party tooling and has a lot of useful controls, behaviors and converters. It is documented extensively on the codeproject and UniDock package that we use for docking depends on it.

Playground/Plugins/Services/MockAuthentication/MockAuthentication.csproj
nick-polyak-ms: We can probably add some checks, but currently, if the folder does not exist, it would get created by the command. Also can we do it in a separate Jira - small issue but might take a bit of time.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/DispatcherAdapter.cs
nick-polyak-ms: IoCy is better than MEF because it is built with MEF in mind. It does not have MEF complexity, it is multiplatform, tested on multiple platforms, open source and well documented on the codeproject. So, I prefer to keep it at least for now.

nick-polyak-ms:
> In my opinion whole "Playground" and "Prototypes" folder structure goes against the idea of using a version control system. If we want to focus efforts on a certain topic, we should create a branch or fork for that effort instead of putting it in the main branch. Basically the whole "common" vs "playground" separation with the idea of putting only stable code in the "common" part is, in reality, a second layer of merging and PRs built upon the already existing system. We should use the existing versioning system to solve code management, not a folder structure.

I kinds of agree about Playground - we can try to find a better name for it. 
Prototypes are necessary for designing the features fast without affecting other code and then moving the re-usuable features into the main framework.
Also after the feature is moved into re-usable class library, the prototype can be used for testing this feature. 
In a sense prototype for visual project is the same as test for non-visual project under Test Driven Development (TDD).

nick-polyak-ms:
> I gave some thought to the topic of "shell" and "core" separation, and I highly disagree with the statement this is an abstraction we can make later. This separation is one of the main features of ComposeUI, and thus, it has to be represented from the start in the main branch. Core code shouldn't depend on any UI framework, that's the job of the replacable shell. At the moment it's way too soon to choose any DI container technology. I'm not even convinced it's something we'll ever need in the main compose project. Configurations where the same thing can have multiple implementations and compositions etc is just asking for trouble with a DI configuration. Using a DI container should be for convenience, not for the sake of using a DI container.

There are two issues mixed here
I agree about the folder structure - I'll move Visual Utilities projects under a different folder - Shell so that Core will be reusable and project under Shell will be replaceable by various implementations of the shell. 

I disagree about the DI. DI, when properly implemented - makes it easier to test, mock and replace the functionality. On top of that, I assume there will be some plugins running within the Shell process and the DI facilitates implementing them.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/DispatcherAdapter.cs
nick-polyak-ms: > What's this attribute?

This attribute means that the ISyncContext is resolved to DispatcherAdapter by the Inversion of Control container.

src/Common/MorganStanley.ComposeUI.Common.BasicModels/MorganStanley.ComposeUI.Common.BasicModels.csproj
nick-polyak-ms: I'd like Core to be a folder containing re-usable and UI framework independent projects, so let us come up with a different name for the project containing ViewModel basic functionality.

src/Common/MorganStanley.ComposeUI.Common.Interfaces/MorganStanley.ComposeUI.Common.Interfaces.csproj
nick-polyak-ms: again, see my comment above.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/MorganStanley.ComposeUI.Common.VisualUtils.csproj
nick-polyak-ms: Yes, correct, I'll rename it to make sure that it is clearly marked as Avalonia project.

nick-polyak-ms:
> Please remove junk files like `src/Directory.Build.targets_SAVE`

This is the old file - did not want to remove it completely, but the code would not compile with it. I guess I'll remove it now.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/MorganStanley.ComposeUI.Common.VisualUtils.csproj
nick-polyak-ms: The VisualUtils will be renamed to underscore that it is an Avalonia implementation.

src/Common/MorganStanley.ComposeUI.Common.Interfaces/MorganStanley.ComposeUI.Common.Interfaces.csproj
nick-polyak-ms: MorganStanley.ComposeUI.Core functionality should be moved under BasicModels project under a different Jira.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/DispatcherAdapter.cs
ztanczos: > This attribute means that the ISyncContext is resolved to DispatcherAdapter by the Inversion of Control container and it is resolved as a Singleton.

Can you please point me to the documentation for this DI container? I don't want to make premature comments without understanding the basics but to me having an `Implements` attribute sounds weird: if a class implements an interface then it should simply implement it, the abstraction-implementation mappings should live in the Composition Root.

nick-polyak-ms: here is some documentation https://www.codeproject.com/Articles/1244087/Introducing-IoCy-Simple-and-Powerful-IoC-Container. It is my own container, so I am a bit partial. The good point through is that it is very small and every software guy or girl can become an expert in it within two hours. 
I am very open to other popular containers, but I think that Microsoft DependencyInjection container misses important functionality. 
ImplementsAttribute notifies that it should map the implementation type to the interface or base class mentioned as the first argument. So when the class is added to the container (especially if it is loaded dynamically) it will map into ISyncContext interface as a singleton and one will be able to obtain the object of the class by using 
`ISyncContext syncContext = container.Resolve<ISyncContext>();`

nick-polyak-ms:
> Nitpick: you renamed Directory.Build.targets to Directory.Build.targets_SAVE - was that intentional?

It was not building without the renaming. I am going to remove the file altogether as Marton suggested.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/MorganStanley.ComposeUI.Common.VisualUtils.csproj
nick-polyak-ms: renamed it to MorganStanley.ComposeUI.Avalonia.VisualUtils and placed it within Visuals/Avalonia physical and solution folders.

src/Common/MorganStanley.ComposeUI.Common.BasicModels/MorganStanley.ComposeUI.Common.BasicModels.csproj
nick-polyak-ms: For now I left it as it is, but placed it within Core folder

nick-polyak-ms: Open for renaming suggestions as part of another Jira.

nick-polyak-ms:
> Please remove junk files like `src/Directory.Build.targets_SAVE`

Done

nick-polyak-ms:
> The folder structure is inconsistent. Bare in mind that we will have javascript/node projects as well. Under `src` we should have stuff like `core` (contains the .NET solutions providing core functionality and default plugin implementations), `avalonia-shell` (projects for the Avalonia-based shell), `javascript-binding` (npm packages for JavaScript modules), etc.

I did it slightly differently but in the spirit of the comment:
currently under Common I have 

- Core folder containing non-visual re-usuable projects
- Visuals\Avalonia folder containing the ...VisualUtils avalonia projects
- Plugins folder
- javascrip-binding folder

Similar folder/project structure is under Tryouts (former Playground) folder.

src/Common/MorganStanley.ComposeUI.Common.Utilities/MorganStanley.ComposeUI.Common.Utilities.csproj
nick-polyak-ms: I still have to keep a refernce to NP.Utilities, until we resolve our disagreements about an IoC framework. I'll remove this as part of IoC Jira https://github.com/morganstanley/ComposeUI/issues/68.

Playground/Plugins/Services/MockAuthentication/MockAuthentication.csproj
nick-polyak-ms: As I stated before I'll open another jira for testing that the path exists, but for now I'll resolve this issue.

src/Common/MorganStanley.ComposeUI.Common.VisualUtils/MorganStanley.ComposeUI.Common.VisualUtils.csproj
nick-polyak-ms: the avalonia related reusable projects are now in their own folders Visuals/Avalonia and have 'Avalonia' as part of their names. Resolving this issue.

nick-polyak-ms:
destroyed the code by mistake, use https://github.com/morganstanley/ComposeUI/pull/69 pull request instead.

PR #55, opened 2022-07-08
Prototypes/hackday1

kruplm:


BalassaMarton (approved):

src/prototype-web/chart/src/chart.js
```diff
@@ -0,0 +1,51 @@
+import Chart from 'highcharts/es-modules/Core/Chart/Chart.js';
```
ZKRobi, line 1: Clarify on highcharts licencing please
This is not an open source software, so I'm not quite of it's usage here.

PR #58, opened 2022-07-11
Prototypes/hackday1 - Pub/Sub/Unsub/Invoke/RegisterService works now

Sly1024:


Sly1024:
I haven't added much error checking, so if the message format is incorrect it might fail, but it can process the messages coming from `ComposeUI.Messaging.Server`.

Sly1024:
I also changed the output format. It now contains both ES6 modules (+d.ts files), and a `bundle.umd.js` for loading it directly in a browser.

ZKRobi (approved):

PR #59, opened 2022-07-13
Prototypes/hackday1 host communication integration

ZKRobi:
WPF host for handling a basic application lifecycle
Initial integration with message router


src/MorganStanley.ComposeUI.Host/MainWindow.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        InitializeComponent();
+        _apps.Add(new HelloWorldApp());
+        _apps.Add(new WebApplication());
+        _moduleLoader = new ModuleLoader(_apps, _communicationModule);
+    }
+
+    private void DisplayApps()
+    {
+        _apps[0].Render(App1);
```
BalassaMarton, line 46: I still don't get why it's hard-coded. `foreach` on a list is cheap enough.

src/MorganStanley.ComposeUI.Host/Modules/CommunicationModule.cs
```diff
@@ -0,0 +1,74 @@
...
+    Process? _process;
+    private bool _disposed;
+
+    public async Task Initialize(IMessageRouter messageRouter)
+    {
+        _process = new Process();
+        _process.StartInfo.UseShellExecute = false;
+        //_process.StartInfo.CreateNoWindow = true;
+
+        var location = Path.GetFullPath("..\\..\\..\\ComposeUI.Messaging.Server\\bin\\Debug\\ComposeUI.Messaging.Server.exe");
```
BalassaMarton, line 35: This could point to the publish output directory instead. Or make it a command line argument.

src/MorganStanley.ComposeUI.Host/MainWindow.xaml.cs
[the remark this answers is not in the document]
ZKRobi: The aim of this current state is to display specifically 2 apps. I already have an improvement in the works that is not the scope of this PR. It doesn't make sense to generalize this code just to throw it out.

src/MorganStanley.ComposeUI.Host/Modules/CommunicationModule.cs
ZKRobi: I would keep it this way for this stage, as it works "out of the box" for the developer.
I think we should discuss how to arrange the output of the different pieces and how to do a better integration for the next step.

kruplm: LGTM for this stage. I saw it working.

PR #61, opened 2022-07-17
ADR - Built-in APIs should be built on top of the Message Router

BalassaMarton:


BalassaMarton:
Created this PR mainly for the sake of discussion. Is there a downside to reducing the endpoints to those needed by the Message Router?

architecture/adr-010-built-in-apis.md
```diff
@@ -0,0 +1,39 @@
...
+## Context
+
+ComposeUI as a framework will provide a set of standard APIs to modules. 
+One of these standard APIs is the pub/sub/invoke functionality of the Message Router.
+Modules will communicate and provide services for each other via the Message Router.
+However, there might be other framework services that we provide, such as login,
+local/profile storage, telemetry, etc. These APIs need client libraries for all the
+languages and technologies we support, including at minimum JavaScript and .NET.
+Since ComposeUI modules can run in their own process, we have to implement cross-process
+communication and build our client libraries around technologies like WebSocket and gRPC
```
mgasca, line 12: Can we use the term "communication protocols" instead of "technologies"?

mgasca:
Hi all, I'm going to be adding some feedback on this, some of which may seem nit-picky.  However, it's good to keep in mind that ADRs don't just serve as implementation notes but (along with the Architecture Decision Log) also as historical documentation used to understand a system.  This is used by contributors exploring something new to them, absolutely new contributors, and external parties alike.  

Being clear and unambiguous is of utmost important, the value of an ADR is not just the decision itself but also the language used.  This is similar to how the value in Patterns is not just the architectural structure or interactions illustrated and described, but in ensuring we speak about it in a common language.  We can also look to the definition and usage of a Ubiquitous Language to see the value of clear, non-ambiguous, shared language.

Given the potential future readers of our ADRs I think we should set the bar at clarity for external readers, as they have the least a priori knowledge of the terms, technologies and internals that existing contributors will have.

psmulovics (approved):

architecture/adr-010-built-in-apis.md
```diff
@@ -0,0 +1,39 @@
...
+adding another port requires us to add all endpoints on all platforms; and supporting a new platform requires us
+to implement all the ports and endpoints on that platform.
+
+## Decision
+
+To reduce the overhead of extending the framework with more ports and APIs, built-in APIs
+will all go through the Message Router, effectively using it as a general-purpose RPC solution.
+Adding a new API will consist of defining a set of new request/response types.
+Adding a new port will only require us to implement communication with the Message Router.
+Supporting one more platform will require a new client library that can connect to
```
mgasca, line 26: Supporting one more platform will only require a new Message Router client library for that platform that can communicate over all the required ports.

architecture/adr-010-built-in-apis.md
```diff
@@ -0,0 +1,39 @@
...
+with the Message Router).
+
+## Status
+
+Draft
+
+## Consequences
+
+This decision affects the design of our APIs. We will have a message-based architecture 
+where APIs are defined in terms of request-response pairs instead of service objects.
```
mgasca, line 37: "where APIs are defined internally interns of request-response pairs between the API client library and the API implementation within a Module or Compose core."  Otherwise it sounds like the API itself is defined by it, but that's not correct.  The API definition is the public function definitions.  Those don't have to be defined in request/response pairs.

architecture/adr-010-built-in-apis.md
```diff
@@ -0,0 +1,39 @@
+# Architecture Decision Record: Built-in APIs should be built on top of the Message Router
+
+## Context
+
+ComposeUI as a framework will provide a set of standard APIs to modules. 
+One of these standard APIs is the pub/sub/invoke functionality of the Message Router.
+Modules will communicate and provide services for each other via the Message Router.
+However, there might be other framework services that we provide, such as login,
```
mgasca, line 8: "there might be other framework service APIs that Compose provides"

Suggest adding API after service for consistency.  Also, suggest "Compose" instead of "we", I feel it's more clear

mgasca (commented):
Have taken the time to do a more full review of the ADR

BalassaMarton:
I've made some changes, please read and comment by eow before I merge it.

ztanczos (dismissed):

PR #70, opened 2022-07-26
Discussion for ADR about process ownership

ZKRobi:
I prepared an initial ADR suggestion about process ownership in Compose. This is a first version that I'd like to give additional thought, so I only created it as a draft PR for now, but feel free to add your insights.

### DIFF
--- a/architecture/adr-011-process-ownership.md
+++ b/architecture/adr-011-process-ownership.md
@@ -1,0 +1,26 @@
+# Architecture Decision Suggestion: Process Ownership
+
+## Context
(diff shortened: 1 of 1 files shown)


psmulovics:
I think the ADR should still contain the ideas we vetted and decided against?

ZKRobi:
> I think the ADR should still contain the ideas we vetted and decided against?

I added a considerations section where I summarized that other options were different levels of giving more direct control of processes to the shell and modules, and described reasons against it.

kruplm (dismissed):

lilla28 (approved):

Issue #73, opened 2022-08-09
Create Windows only functionality allowing one Visual process not to block another processes via visual message queue


nick-polyak-ms:
The idea is that we shall use DWM (Desktop Window Manager) to display windows for different processes similar to WARP. It has been proven to work.

PR #79, opened 2022-08-19
POCs for hosting DWM thumbs instead of the real windows and Docking them in Avalonia

nick-polyak-ms:
This is useful, because DWM thumbs will not lock the rest of the application if hanging e.g. due to a long sleep.
https://github.com/morganstanley/ComposeUI/issues/73

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -14,6 +14,22 @@ jobs:
   build:
 
     runs-on: windows-latest
+    strategy:
+      matrix:
+        dotnet-version: [ '6.0.x' ]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
+    - name: Setup .NET Core SDK ${{ matrix.dotnet-version }}
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version: ${{ matrix.dotnet-version }}
+    - name: Install dependencies
+      run: dotnet restore
+      working-directory: src
+    - name: Build
+      run: dotnet build --configuration Release --no-restore
+      working-directory: src
+    - name: Test
+      run: dotnet test --no-restore --verbosity normal
+      working-directory: src


Tryouts/Plugins/ApplicationPlugins/DwmSourceWpfApp/App.xaml.cs
```diff
@@ -0,0 +1,56 @@
...
+            Container.InjectPluginsFromSubFolders("Plugins/Services");
+            Container.CompleteConfiguration();
+            _subscriptionClient = Container.Resolve<ISubscriptionClient>();
+        }
+
+        protected override async void OnStartup(StartupEventArgs e)
+        {
+            base.OnStartup(e);
+
+            await _subscriptionClient.Connect(CommunicationsConstants.MachineName, CommunicationsConstants.Port);
```
ztanczos, line 42: Does this mean that the application hosted here needs to 'connect' to something to be able to displayed as a DWM thumb?

Tryouts/Visuals/WindowsOnly/DwmUtilsForAvalonia/Dwm/DwmVisualProxy.cs
```diff
@@ -0,0 +1,374 @@
...
+
+            // WinApi.SendMessage(visualHandle, 0x009B, IntPtr.Zero, IntPtr.Zero);
+
+            RegisterMouseEvents();
+            RegisterKeyboardEvents();
+        }
+
+        private async void OnSizeChanged()
+        {
+            _subscriptionClient.Publish(Topic.WindowSize, new WindowCoordMessage { WindowId = (long)visualHandle, X = (int)Bounds.Width, Y = (int)Bounds.Height});
```
ztanczos, line 74: Do you need to 'await' this operation?

ztanczos:
There are lots of commented out code lines in this pull request. Why are they commented out? Are they there to illustrate / document certain things for other developers? If so then I think you should also add some context for each comment, not just code-snippets.
If they were just left there by accident then I'd suggest removing them.

Tryouts/Visuals/WindowsOnly/WinDwmSourceUtils/HwndMouseInputProviderEx.cs
```diff
@@ -0,0 +1,827 @@
+﻿/// ********************************************************************************************************
+///
+/// Morgan Stanley makes this available to you under the Apache License, Version 2.0 (the "License").
```
ztanczos, line 3: There are code snippets in here which are from Microsoft - can you please clarify which license covers those?

Tryouts/Plugins/Services/CommunicationsServices/DwmUtils/DwmVisual.cs
```diff
@@ -0,0 +1,273 @@
...
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Data;
+using System.Windows.Input;
+using System.Windows.Interop;
+using System.Windows.Media;
+using Subscriptions;
+using MorganStanley.ComposeUI.Tryouts.Visuals.WindowsOnly.Win32Calls;
+
+namespace MorganStanley.ComposeUI.Services.CommunicationsServices.DwmUtils
```
ztanczos, line 16: What does DWM have to do with Communications Services?

Tryouts/Plugins/Services/CommunicationsServices/DwmUtils/HwndMouseInputProviderEx.cs
```diff
@@ -0,0 +1,816 @@
...
+using System.Runtime.InteropServices;
+using System.Windows;
+using System.Windows.Input;
+using System.Windows.Interop;
+using System.Windows.Media;
+using static MorganStanley.ComposeUI.Tryouts.Visuals.WindowsOnly.Win32Calls.WinApi;
+
+namespace MorganStanley.ComposeUI.Services.CommunicationsServices.DwmUtils
+{
+    class HwndMouseInputProviderEx
```
ztanczos, line 15: This file has been added 3x with this PR

Tryouts/Plugins/ApplicationPlugins/DwmSourceWpfApp/App.xaml.cs
nick-polyak-ms: Connection is needed for communications between the shell and the WPF process behind the DWM thumb, not for the initial display. In this respect it is close to how it was implemented in MSDesktop.

Tryouts/Visuals/WindowsOnly/DwmUtilsForAvalonia/Dwm/DwmVisualProxy.cs
nick-polyak-ms: No need to await on this operation, since it does not return anything - just publishes a message.

ztanczos: You can use a discard to be explicit about the intention that this is just fire-and-forget, no need to await the Task which was returned.
```
_ = _subscriptionClient.Publish(...)
```

Tryouts/Visuals/WindowsOnly/WinDwmSourceUtils/HwndMouseInputProviderEx.cs
nick-polyak-ms: Actually it is not Microsoft code - it is our own code from MSDesktop. So I assume we can use it in ComposeUI also (which is a replacement for MSDesktop).

Tryouts/Plugins/Services/CommunicationsServices/DwmUtils/DwmVisual.cs
ztanczos: I understand that the DWM implementation would rely on communication but here it seems DwmUtils were added directly into the communication service. Shouldn't the two be separated?

PR #83, opened 2022-08-23
Fix the CI build

ZKRobi:
fix provided by @nick-polyak-ms


ztanczos (approved):

ztanczos:
Apologies - I misread the package's name being added. Can you please revert this commit and only add dependency to Avalonia?

PR #86, opened 2022-08-30
Visual functionality for using modules together with UniDock layout  (including saving and restoring the layout)

nick-polyak-ms:
Also created a prototype for demonstrating the functionality

### DIFF
--- a/Tryouts/Core/BasicModels/Modules/ModuleViewModel.cs
+++ b/Tryouts/Core/BasicModels/Modules/ModuleViewModel.cs
@@ -1,0 +1,51 @@
+﻿/// ********************************************************************************************************
+///
+/// Morgan Stanley makes this available to you under the Apache License, Version 2.0 (the "License").
+/// You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+/// See the NOTICE file distributed with this work for additional information regarding copyright ownership.
+/// Unless required by applicable law or agreed to in writing, software distributed under the License
+/// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and limitations under the License.
+/// 
+/// ********************************************************************************************************
+
+using MorganStanley.ComposeUI.Tryouts.Core.Abstractions.Modules;
+
+namespace MorganStanley.ComposeUI.Tryouts.Core.BasicModels.Modules
+{
+    public class ModuleViewModel
+    {
+        public ModuleManifest Manifest { get; init;  }
+
+        public ModuleViewModel(ModuleManifest manifest)
+        {
+            Manifest = manifest;
+        }
+
+        public ModuleViewModel
+        (
+            string name, 
+            StartupType startupType, 
+            string uiType, 
+            string pathOrUrl) 
+            : 
+            this(new ModuleManifest { Name = name, StartupType = startupType, UIType = uiType })
+        {
+            if (uiType == UIType.Window)
+            {
+                Manifest.Path = pathOrUrl;
+            }    
+            else
+            {
+                Manifest.Url = pathOrUrl;
+            }
+        }
+
+        public event Action<ModuleManifest>? LaunchEvent;
+
+        public void Launch()
+        { 
+            LaunchEvent?.Invoke(Manifest);
+        }
+    }
+}
--- a/Tryouts/Core/BasicModels/Modules/ProcessDockLayoutBehavior.cs
+++ b/Tryouts/Core/BasicModels/Modules/ProcessDockLayoutBehavior.cs
@@ -1,0 +1,234 @@
+﻿/// ********************************************************************************************************
+///
+/// Morgan Stanley makes this available to you under the Apache License, Version 2.0 (the "License").
+/// You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+/// See the NOTICE file distributed with this work for additional information regarding copyright ownership.
+/// Unless required by applicable law or agreed to in writing, software distributed under the License
+/// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and limitations under the License.
+/// 
+/// ********************************************************************************************************
+
+using NP.Avalonia.UniDockService;
+using NP.Concepts.Behaviors;
+using System;
+using System.Reactive.Linq;
+
+namespace MorganStanley.ComposeUI.Tryouts.Core.BasicModels.Modules
+{
+    public class ProcessDockLayoutBehavior
+    {
+        IUniDockService _uniDockService;
+        ProcessesViewModel _processesViewModel;
+        string _dockSerializationFileName;
+        string _vmSerializationFileName;
+        string _mainGroupDockId;
+        string _contentTemplateResourceKey;
+        string _headerContentTemplateResourceKey;
+
+        int _newDockId = 0;
+
+        SynchronizationContext _synchronizationContext;
+
+        IDisposable? _uniDockVmBehavior;
+        IDisposable? _processesWithWindowsBehavior;
+
+        public ProcessDockLayoutBehavior
+        (
+            IUniDockService uniDockService,
+            ProcessesViewModel processesViewModel,
+            string dockSerializationFileName,
+            string vmSerializationFileName,
+            string mainGroupDockId,
+            string contentTemplateResourceKey,
+            string headerContentTemplateResourceKey
+        )
+        {
+            _uniDockService = uniDockService;
+            _processesViewModel = processesViewModel;
+            _dockSerializationFileName = dockSerializationFileName;
+            _vmSerializationFileName = vmSerializationFileName;
+            _mainGroupDockId = mainGroupDockId;
+            _contentTemplateResourceKey = contentTemplateResourceKey;
+            _headerContentTemplateResourceKey = headerContentTemplateResourceKey;
+
+#pragma warning disable CS8974 // Converting method group to non-delegate type
+            _processesWithWindowsBehavior = 
+                _processesViewModel.ProcessesWithWindows.AddBehavior(OnProcessWithWindowAdded, OnProcessWithWindowRemoved);
+#pragma warning restore CS8974 // Converting method group to non-delegate type
+
+            _synchronizationContext = SynchronizationContext.Current!;
+
+            ResetBehavior();
+        }
+
+        private void ResetBehavior()
+        {
+            _uniDockVmBehavior?.Dispose();
+
+            _uniDockVmBehavior = _uniDockService.DockItemsViewModels.AddBehavior(OnVmAdded, OnVmRemoved);
+        }
+
+        private void OnVmAdded(DockItemViewModelBase vm)
+        {
+
+        }
+
+        private void OnVmRemoved(DockItemViewModelBase vm)
+        {
+            if (_isLoading)
+            {
+                return;
+            }
+            DockItemViewModel<ProcessData> processVm = 
+                (DockItemViewModel<ProcessData>)vm;
+
+            Guid instanceId = 
+                processVm.TheVM!.InstanceId;
+
+            _processesViewModel.StopProcess(instanceId);
+        }
+
+        public void Save()
+        {
+            _uniDockService.SaveToFile(_dockSerializationFileName);
+
+            _uniDockService.SaveViewModelsToFile(_vmSerializationFileName);
+        }
+
+        bool _isLoading = false;
+        public void Load()
+        {
+            _isLoading = true;
+
+            try
+            {
+                _uniDockVmBehavior?.Dispose();
+                _uniDockVmBehavior = null;
+                _processesWithWindowsBehavior?.Dispose();
+                _processesWithWindowsBehavior = null;
+                _uniDockService.RestoreFromFile(_dockSerializationFileName, false);
+                _uniDockService.RestoreViewModelsFromFile(_vmSerializationFileName, typeof(DockItemViewModel<ProcessData>));
+
+                _newDockId =
+                    _uniDockService
+                        .DockItemsViewModels
+                        .Cast<DockItemViewModel<ProcessData>>()
+                        .Max(dockItemVm => dockItemVm.TheVM!.WindowNumber);
+
+                var allInstanceIds =
+                    _uniDockService
+                        .DockItemsViewModels
+                        .Cast<DockItemViewModel<ProcessData>>()
+                        .Select(vm => vm.TheVM.InstanceId)
+                        .Distinct()
+                        .ToList();
+
+                _processesViewModel.ClearProcessesThatDoNotMatchIds(allInstanceIds);
+
+
+                _processesWithWindowsBehavior =
+                    _processesViewModel.ProcessesWithWindows.AddBehavior(OnProcessWithWindowAdded, OnProcessWithWindowRemoved);
+
+                foreach (DockItemViewModel<ProcessData> vm in _uniDockService.DockItemsViewModels)
+                {
+                    string processName = vm.TheVM!.ProcessName!;
+
+                    Guid instanceId = vm.TheVM.InstanceId;
+
+                    bool launchedProcess = _processesViewModel!.LaunchProcess(processName, instanceId);
+
+                    if (!launchedProcess)
+                    {
+                        SingleProcessViewModel? singleProcessViewModel = _processesViewModel?.FindRunningProcess(instanceId);
+
+                        if (singleProcessViewModel != null)
+                        {
+                            OnProcessWithWindowAdded(singleProcessViewModel);
+                        }
+                    }
+                }
+                ResetBehavior();
+            }
+            finally
+            {
+                _isLoading = false;
+            }
+        }
+
+
+        private void OnProcessWithWindowAdded(SingleProcessViewModel processViewModel)
+        {
+            _synchronizationContext.Send
+            (
+                (_) =>
+                {
+                    Guid instanceId = processViewModel.InstanceId;
+
+                    var existingVm =
+                        _uniDockService
+                            .DockItemsViewModels
+                            .Cast<DockItemViewModel<ProcessData>>()
+                            .FirstOrDefault(dockItemVm => dockItemVm.TheVM!.InstanceId == instanceId);
+
+                    if (existingVm != null)
+                    {
+                        existingVm.TheVM!.ProcessMainWindowHandle = processViewModel.ProcessMainWindowHandle;
+                    }
+                    else
+                    {
+                        _newDockId++;
+                        var dockItemViewModel = new DockItemViewModel<ProcessData>
+                        {
+                            DockId = _newDockId.ToString(),
+                            DefaultDockGroupId = _mainGroupDockId,
+                            DefaultDockOrderInGroup = _newDockId,
+                            ContentTemplateResourceKey = _contentTemplateResourceKey,
+                            HeaderContentTemplateResourceKey = _headerContentTemplateResourceKey,
+                            IsSelected = true,
+                            IsActive = true,
+                            IsPredefined = false,
+                        };
+
+                        dockItemViewModel.TheVM = new ProcessData
+                        {
+                            InstanceId = processViewModel.InstanceId,
+                            ProcessName = processViewModel.ProcessName,
+                            WindowNumber = _newDockId,
+                            ProcessMainWindowHandle = processViewModel.ProcessMainWindowHandle
+                        };
+                        _uniDockService!.DockItemsViewModels.Add(dockItemViewModel);
+                    }
+                },
+                null
+            );
+
+        }
+
+        private void OnProcessWithWindowRemoved(SingleProcessViewModel processViewModel)
+        {
+            if (_isLoading)
+            {
+                return;
+            }
+            _synchronizationContext.Send((_) =>
+            {
+                var dockVm =
+                    _uniDockService
+                    .DockItemsViewModels
+                    .Cast<DockItemViewModel<ProcessData>>()
+                    .FirstOrDefault(vm => vm.TheVM!.InstanceId == processViewModel.InstanceId);
+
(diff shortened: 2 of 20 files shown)


Tryouts/Core/BasicModels/Modules/ModuleViewModel.cs
```diff
@@ -0,0 +1,25 @@
...
+/// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and limitations under the License.
+/// 
+/// ********************************************************************************************************
+
+using MorganStanley.ComposeUI.Tryouts.Core.Abstractions.Modules;
+
+namespace MorganStanley.ComposeUI.Tryouts.Core.BasicModels.Modules
+{
+    public class ModuleViewModel : ModuleManifest
```
ztanczos, line 16: Why did you choose inheritance? To me it feels wrong that a view model inherits from a module manifest, I'd rather prefer composition.

Tryouts/Core/BasicModels/Modules/ProcessDockLayoutBehavior.cs
```diff
@@ -0,0 +1,234 @@
...
+/// See the License for the specific language governing permissions and limitations under the License.
+/// 
+/// ********************************************************************************************************
+
+using NP.Avalonia.UniDockService;
+using NP.Concepts.Behaviors;
+using System;
+using System.Reactive.Linq;
+
+namespace MorganStanley.ComposeUI.Tryouts.Core.BasicModels.Modules
```
ztanczos, line 17: I'm getting confused by this namespace: do we have models, modules, view models, DTOs, or UI behaviors here?

Tryouts/Core/BasicModels/Modules/ProcessesViewModel.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        public IEnumerable<SingleProcessViewModel> Processes { get; } =
+            new ObservableCollection<SingleProcessViewModel>();
+
+
+        public IEnumerable<SingleProcessViewModel> ProcessesWithWindows { get; } =
+            new ObservableCollection<SingleProcessViewModel>();
+
+        private IDisposable _modulesBehavior;
+        private IDisposable _processesBehavior;
```
ztanczos, line 34: Who's responsible for disposing these IDisposables?

Tryouts/Prototypes/ModulesDockingPrototype/MorganStanley.ComposeUI.Prototypes.ModulesDockingPrototype.csproj
```diff
@@ -0,0 +1,48 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <!--This helps with theme dll-s trimming.
+	If you will publish your application in self-contained mode with p:PublishTrimmed=true and it will use Fluent theme Default theme will be trimmed from the output and vice versa.
+	https://github.com/AvaloniaUI/Avalonia/issues/5593 -->
+    <TrimmableAssembly Include="Avalonia.Themes.Fluent" />
+    <TrimmableAssembly Include="Avalonia.Themes.Default" />
+  </ItemGroup>
+  <ItemGroup>
+    <!--Condition below is needed to remove Avalonia.Diagnostics package from build output in Release configuration.-->
```
ztanczos, line 35: I don't see any condition below

Tryouts/Core/BasicModels/Modules/ProcessesContainer.cs
```diff
@@ -0,0 +1,151 @@
...
+/// 
+/// ********************************************************************************************************
+
+using MorganStanley.ComposeUI.Tryouts.Core.Abstractions.Modules;
+using NP.Concepts.Behaviors;
+using System.Collections.ObjectModel;
+
+namespace MorganStanley.ComposeUI.Tryouts.Core.BasicModels.Modules
+{
+    public class ProcessesContainer
```
ztanczos, line 18: Where/how is this class used? I don't see any references to it.

Tryouts/Core/BasicModels/Modules/ProcessDockLayoutBehavior.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            if (_isLoading)
+            {
+                return;
+            }
+            DockItemViewModel<ProcessData> processVm = 
+                (DockItemViewModel<ProcessData>)vm;
+
+            Guid instanceId = 
+                processVm.TheVM!.InstanceId;
```
ztanczos, line 87: What's "TheVM"?

Tryouts/Core/BasicModels/Modules/ProcessDockLayoutBehavior.cs
nick-polyak-ms: BasicModels is actually not for DTOs - but for the re-usable ViewModels required by the framework. In general, on the UI side it is a good practice to avoid DTOs (plain models) as much as possible dealing straight with ViewModels (that contain also DTO information). In the case above (which I now changed to composition) ModuleManifest is a Model (DTO), while ModuleViewModel is the view model, but in many cases, View Models do not have to have a separate Model counterpart. 

Modules at the end specify that these View Models are dealing with the Module functionality. Yes, unfortunately, Module  and Model are kind of similar in spelling an pronunciation, but there is no real similarity between them.

Tryouts/Core/BasicModels/Modules/ProcessesViewModel.cs
nick-polyak-ms: They are not disposed, they stay there for the duration of the main process, to prevent the corresponding behaviors from disappearing due to a lack of reference.

Tryouts/Core/BasicModels/Modules/ProcessesContainer.cs
nick-polyak-ms: very good point - it was one of the working sketches that got checked in by mistake.

Tryouts/Core/BasicModels/Modules/ProcessDockLayoutBehavior.cs
nick-polyak-ms: this is the actual View Model of type ProcessData. This is part of the UniDock framework.

Tryouts/Core/BasicModels/Modules/ModuleViewModel.cs
nick-polyak-ms: replaced inheritance with composition

Tryouts/Core/BasicModels/Modules/ProcessesViewModel.cs
```diff
@@ -0,0 +1,208 @@
...
+        private void OnProcessesAdded(SingleProcessViewModel process)
+        {
+            process.StopEvent += OnStopProcess;
+            process.StartedEvent += Process_StartedEvent;
+            process.StoppedEvent += Process_StoppedEvent;
+        }
+
+        private void Process_StoppedEvent(SingleProcessViewModel process)
+        {
+            ((IList<SingleProcessViewModel>)ProcessesWithWindows).Remove(process);
```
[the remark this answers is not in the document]
nick-polyak-ms: added private _processes and _prossessesWithWindows of type `Observable<SingleProcessViewModel>` so that I could remove the castings. 
The properties Processes and ProcessesWithWindows are still of `IEnumerable<SingleProcessViewModel>` so that no client of the class would be able to modify them.

Tryouts/Core/BasicModels/Modules/ProcessesContainer.cs
nick-polyak-ms: Removed the unneeded file

Tryouts/Prototypes/ModulesDockingPrototype/MorganStanley.ComposeUI.Prototypes.ModulesDockingPrototype.csproj
nick-polyak-ms: Removed the outdated comment

Tryouts/Visuals/Avalonia/VisualUtils/NativeHostBase.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                style = (style & ~((uint)WS_POPUP | (uint)WS_CAPTION | (uint)WS_THICKFRAME | (uint)WS_MINIMIZEBOX | (uint)WS_MAXIMIZEBOX | (uint)WS_SYSMENU)) | (uint)WS_CHILD;
+
+                SetWindowLongPtr(new HandleRef(null, WindowHandle), (int)GWL_STYLE, (IntPtr)style);
+            }
+
+            // force refreshing the handle
+            MethodInfo methodInfo =
+                typeof(NativeControlHost)
+                    .GetMethod("DestroyNativeControl", BindingFlags.Instance | BindingFlags.NonPublic);
```
ztanczos, line 60: This seems risky: we'd be relying on the internal implementation of Avalonia.

Tryouts/Visuals/Avalonia/VisualUtils/NativeHostBase.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                return base.CreateNativeControlCore(parent);
+            }
+
+        }
+
+        protected override void DestroyNativeControlCore(IPlatformHandle control)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
```
ztanczos, line 103: Is this intentionally left blank?

Tryouts/Prototypes/ModulesDockingPrototype/MorganStanley.ComposeUI.Prototypes.ModulesDockingPrototype.csproj
```diff
@@ -0,0 +1,47 @@
...
+Microsoft Visual Studio Solution File, Format Version 12.00
+
+********************************************************************************************************
+-->
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <!--Avalonia doesen't support TrimMode=link currently,but we are working on that https://github.com/AvaloniaUI/Avalonia/issues/6892 -->
```
ztanczos, line 19: This github issue has been closed on 20th Febr

Tryouts/Prototypes/ModulesDockingPrototype/MorganStanley.ComposeUI.Prototypes.ModulesDockingPrototype.csproj
```diff
@@ -0,0 +1,47 @@
...
+    <BuiltInComInteropSupport>true</BuiltInComInteropSupport>
+    <ApplicationManifest>app.manifest</ApplicationManifest>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Remove=".gitignore" />
+  </ItemGroup>
+  <ItemGroup>
+    <!--This helps with theme dll-s trimming.
+	If you will publish your application in self-contained mode with p:PublishTrimmed=true and it will use Fluent theme Default theme will be trimmed from the output and vice versa.
+	https://github.com/AvaloniaUI/Avalonia/issues/5593 -->
```
ztanczos, line 30: This github issue has been closed on 17th May.

Tryouts/Visuals/Avalonia/VisualUtils/WindowBasedNativeHost.cs
```diff
@@ -0,0 +1,25 @@
...
+/// See the NOTICE file distributed with this work for additional information regarding copyright ownership.
+/// Unless required by applicable law or agreed to in writing, software distributed under the License
+/// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and limitations under the License.
+/// 
+/// ********************************************************************************************************
+
+using System;
+
+namespace MorganStanley.ComposeUI.Tryouts.Visuals.Avalonia.VisualUtils
```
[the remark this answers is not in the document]
nick-polyak-ms: 'Visuals' link within the namespace specifies that it has dependencies on Visual packages - Avalonia or WPF. It separates it from Core (which does not depend on any visual packages). 
'Avalonia' link specifies that it is build on top of Avalonia - I expect that there might be also projects starting with `MorganStanley.ComposeUI.Tryouts.Visuals.WPF.`. 
The final link - 'VisualUtils' mean that these are the generic visual utilities, can be used by any project within ComposeUI. We shall probably also have some common Views under Avalonia, containing more specific code that applies only to that view. So I assume that we shall have e.g. project MorganStanley.ComposeUI.Tryouts.Visuals.Avalonia.AuthenticationView. 

If later it turns out that some of the links are superfluous, we can always go back and delete them, also rearranging the code correspondingly - but that would be a large architectural Jira on its own.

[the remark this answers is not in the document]
ztanczos: Yep, my point is exactly that everything under Visuals should be related to visual packages, so I don't expect non-visual utils there hence '*Visual*Utils' is redundant, it could be simply 'Utils'. But as I said it's very minor.

Tryouts/Visuals/Avalonia/VisualUtils/NativeHostBase.cs
nick-polyak-ms: Yes, I simplified the code.

Tryouts/Prototypes/ModulesDockingPrototype/MorganStanley.ComposeUI.Prototypes.ModulesDockingPrototype.csproj
nick-polyak-ms: removed the code

Tryouts/Prototypes/ModulesDockingPrototype/MorganStanley.ComposeUI.Prototypes.ModulesDockingPrototype.csproj
nick-polyak-ms: removed the comment line.

Tryouts/Visuals/Avalonia/VisualUtils/NativeHostBase.cs
nick-polyak-ms: I sent a request to Avalonia to make `NativeControlHost.DestroyNativeControl()` method visible in its derived classes or to provide a better workaround - https://github.com/AvaloniaUI/Avalonia/issues/8875. 
Once they respond, we'll modify our code - I open a Jira to cover it: http://jiraeai.ms.com/jira/projects/COMPOSEUI/issues/COMPOSEUI-40.

Tryouts/Visuals/Avalonia/VisualUtils/WindowBasedNativeHost.cs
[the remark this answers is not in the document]
nick-polyak-ms: Oh, so I misunderstood what you've said. I think it is still better to have VisualUtils not to mix them with non-visual Utilities from the core :)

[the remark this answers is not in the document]
ztanczos: You can't mix them since one is from .Core and the other is from .Visuals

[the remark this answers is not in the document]
nick-polyak-ms: In my experience the last link in the namespace is pretty important so I believe that VisualUtils is better since they leave no doubt to the developers when it is from. I am not averse to change the namespace, but let us do it in a different Jira after the team discussion.

Tryouts/Core/BasicModels/MorganStanley.ComposeUI.Tryouts.Core.BasicModels.csproj
```diff
@@ -18,9 +18,16 @@ See the License for the specific language governing permissions and limitations
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="NP.Avalonia.UniDockService" Version="0.99.9" />
```
ZKRobi, line 22: Why are these packages referenced in Core/BasicModels? To me it very much seems like something specific to the shell implementation

Tryouts/Core/Services/ModulesService/MorganStanley.ComposeUI.Tryouts.Core.Services.ModulesService.csproj
```diff
@@ -28,10 +28,4 @@ See the License for the specific language governing permissions and limitations
     <ProjectReference Include="..\..\..\Core\Abstractions\MorganStanley.ComposeUI.Tryouts.Core.Abstractions.csproj" />
     <ProjectReference Include="..\..\..\Core\Utilities\MorganStanley.ComposeUI.Tryouts.Core.Utilities.csproj" />
   </ItemGroup>
-
-  <Target Name="PostBuild" AfterTargets="PostBuildEvent">
-  	<Exec Command="xcopy &quot;$(OutDir)\$(AssemblyName).dll&quot; &quot;$(SolutionDir)\bin\$(Configuration)\net6.0\Plugins\Services\$(ProjectName)\&quot; /R /Y /I" />
-  	<Exec Command="xcopy &quot;$(OutDir)\$(AssemblyName).pdb&quot; &quot;$(SolutionDir)\bin\$(Configuration)\net6.0\Plugins\Services\$(ProjectName)\&quot; /R /Y /I" />
-  </Target>
-
```
ZKRobi, line 36: I agree with removing these, it seems like a bad idea to copy specifically like this, but make sure it doesn't break other people's work without discussing with them

Tryouts/Core/BasicModels/MorganStanley.ComposeUI.Tryouts.Core.BasicModels.csproj
nick-polyak-ms: BasicModels project by its nature contains the non-visual ViewModels to support the visual projects - the shell (the most important visual project), visual plugins and the prototypes. So, yes, this code still needed to be shared between the prototypes and the shell and perhaps (in the future) some plugins also, so I do think that it belongs here. 

UniDockService, BTW is a very simple - **completely non-visual** interface to UniDock (in case you were worried that it has some visual references).

Tryouts/Core/Services/ModulesService/MorganStanley.ComposeUI.Tryouts.Core.Services.ModulesService.csproj
nick-polyak-ms: These was the left out code from a very brief time when the Modules implementations was a dynamically loaded plugin. It is not needed anymore since the Modules are now statically loaded. 
No other projects will be affected - tested it on my own machine.

Tryouts/Core/BasicModels/MorganStanley.ComposeUI.Tryouts.Core.BasicModels.csproj
nick-polyak-ms: No, neither of the packages have any references to Avalonia and they are not Avalonia specific. Nothing is imposed. 

The behaivors - contain non-visual behaviors - that simplify dealing with observable collections. They allow to specify what will happen when an item becomes part of the collection and when it is removed. If you insist I can remove the reference to behaviors package, but it will considerably expand the code. I'd have to add various handlers to the CollectionChanged event figure out the type of the event (adding, removing etc) and do the corresponding action depending on its type - something the behaviors do.

ztanczos: I know that we're discussing naming again but why is it called NP.**Avalonia.**UniDockService if it's not Avalonia-specific?

ZKRobi: + Something called UniDockService also sounds pretty technology specific to me

nick-polyak-ms: You are absolutely correct. I'll remove Avalonia from the name - it does not belong there. In case UniDock will be implemented by another framework, e.g. WPF, the UniDockService will still be able to be use for it without a single change. Can we still approve it for now and I'll create a Jira - to use the renamed package for it, once it will be available - next Monday.

ZKRobi: But this makes a shell that doesn't want to use UniDock dependent on UniDock specific library
If you think a library that provides generic tools for UniDock, move it to a separate project that's thematic to that piece of technology

nick-polyak-ms: UniDock provides non-visual interface to UniDock. But it does not force anyone to use UniDock. It is a very small set of non-visual code that will not be doing anything back if unidock is not used.

nick-polyak-ms: adding a project to the reusuable statically loaded project should not be done lightly - it is a major architectural change. What I can probably do - seeing so many objections - I can create a dynamically loaded plugin and use it for the shell.

nick-polyak-ms: In order to avoid making BasicModels dependent on UniDockServices and Behaviors packages, I moved most of the classes out of there into a new dynamically loaded plugin, leaving there only simple data and interfaces. BasicModels now do not have references to the libraries (those reference were moved to the new plugin - which is no a shared project and can be easily replaced if needed).  
Please, see if you can approve and merge the project. Thanks

ZKRobi (approved):

PR #95, opened 2022-09-09
Improvements to ModuleLoader

ZKRobi:
Several improvements to the ModuleLoader
Set up sample with messagerouter and sample apps from hackday1 (including DataService)


Tryouts/Core/Services/ModulesService/Hosts/BackgroundModuleHost.cs
```diff
@@ -0,0 +1,44 @@
...
+/// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and limitations under the License.
+/// 
+/// ********************************************************************************************************
+
+using MorganStanley.ComposeUI.Tryouts.Core.Abstractions.Modules;
+
+namespace MorganStanley.ComposeUI.Tryouts.Core.Services.ModulesService.Hosts
+{
+    internal class BackgroundModuleHost : ModuleBase
```
ztanczos, line 16: Shouldn't we also rename ModuleBase to ModuleHostBase?

Tryouts/Core/Services/ModulesService/Runners/ComposeHostedWebApp.cs
```diff
@@ -0,0 +1,55 @@
...
+            _process.StartInfo.ArgumentList.Add(_port.ToString());
+            _process.StartInfo.WorkingDirectory = Path.GetFullPath(_path);
+            _process.Start();
+
+            return Task.Delay(TimeSpan.FromMilliseconds(100));
+        }
+
+        public async Task Stop()
+        {
+            //const string ctrlC = "\x3";
```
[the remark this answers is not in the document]
ZKRobi: Done

Tryouts/Core/Services/ModulesService/Hosts/BackgroundModuleHost.cs
ZKRobi: Fixed inconsistent names related to ModuleHost

Tryouts/Core/Services/ModulesService/Hosts/BackgroundModuleHost.cs
```diff
@@ -0,0 +1,44 @@
...
+    internal class BackgroundModuleHost : ModuleHostBase
+    {
+        private readonly IModuleRunner _runner;
+
+        public BackgroundModuleHost(string name, Guid instanceId, IModuleRunner runner) : base(name, instanceId)
+        {
+            _runner = runner;
+        }
+
+        public override ProcessInfo ProcessInfo => new ProcessInfo(
```
ztanczos, line 25: Nitpick: if this is a property then I wouldn't instantiate a new object every time it is accessed.

ZKRobi: This was brought up in the previous PR as well, ProcessInfo was turned into a struct based on that discussion. The reasoning is that the contents of the class can change, but it should be immutable, and changes should be communicated via the LifeCycle events, not by e.g. monitoring it
I'm open for discussion about alternative solutions :)

nick-polyak-ms:
I am only starting to review it now - so please, give me several days. 

One thing I noticed right away is the folder names under Core/Services. Currently it is:
ComposeUI.Messaging.Client
ComposeUI.Messaging.Core
ComposeUI.Messaging.IntegrationTests
ComposeUI.Messaging.Server
ComposeUI.Messaging.Server.Tests
messaging-web-client

We agreed to shorten the folder names not to have many periods, but to use the full length names for the namespaces and csproj files. So, please, remove ComposeUI. as a prefix here.

ZKRobi:
> I am only starting to review it now - so please, give me several days.
> 
> One thing I noticed right away is the folder names under Tryouts/Core/Services/Messaging. Currently the folder names are: ComposeUI.Messaging.Client ComposeUI.Messaging.Core ComposeUI.Messaging.IntegrationTests ComposeUI.Messaging.Server ComposeUI.Messaging.Server.Tests messaging-web-client
> 
> We agreed to shorten the folder names not to have many periods, but to use the full length names for the namespaces and csproj files. So, please, remove 'ComposeUI.Messaging.' as a prefix here for all folders that have it.

Hi Nick,
This PR is older than that decision, and it already does much with the relocation of folders.
In general a PR should focus on a single goal.
I agree with the renaming, but it is not scope here, and it will be done separately.

nick-polyak-ms:
> 



> > I am only starting to review it now - so please, give me several days.
> > One thing I noticed right away is the folder names under Tryouts/Core/Services/Messaging. Currently the folder names are: ComposeUI.Messaging.Client ComposeUI.Messaging.Core ComposeUI.Messaging.IntegrationTests ComposeUI.Messaging.Server ComposeUI.Messaging.Server.Tests messaging-web-client
> > We agreed to shorten the folder names not to have many periods, but to use the full length names for the namespaces and csproj files. So, please, remove 'ComposeUI.Messaging.' as a prefix here for all folders that have it.
> 
> Hi Nick, This PR is older than that decision, and it already does much with the relocation of folders. In general a PR should focus on a single goal. I agree with the renaming, but it is not scope here, and it will be done separately.

Hey Robi, 
in my humble opinion, PR is not ready until it is move over to the re-usable place and become available for the prototypes and main projects. Yes, I do believe that renaming the folders and other important changes are part of this PR.

nick-polyak-ms:
As we've discussed before -     

    ValueTask<IDisposable> SubscribeAsync(
        string topicName,
        IObserver<RouterMessage> observer,
        CancellationToken cancellationToken = default);

should be returning an IObservable<RoutedMessage> instead of taking in IObserver. It is much easier to deal with IObservable (which is compositional and can be combined filtered aggregated etc using LINQ iterators, than to use IObserver (a dual of IEnumerator - never explicitely used).

ztanczos:
> As we've discussed before -
> 
> ```
> ValueTask<IDisposable> SubscribeAsync(
>     string topicName,
>     IObserver<RouterMessage> observer,
>     CancellationToken cancellationToken = default);
> ```
> 
> should be returning an IObservable instead of taking in an IObserver. It is much easier to deal with IObservable (which is compositional and can be combined filtered aggregated etc using LINQ iterators, than to use IObserver (a dual of IEnumerator - never explicitely used).
> 
> The approximate API of this call should be:
> 
> ```
> (IObservable<RouterMessage> messageStream, ValueTask<IDisposable> disposable) SubscribeAsync(
>     string topicName,
>     CancellationToken cancellationToken = default);
> ```

As we've discussed before - this is existing code, it wasn't introduced with this PR. Feel free to bring up this topic in the next discussion.

BalassaMarton:
> ```cs
> (IObservable<RouterMessage> messageStream, ValueTask<IDisposable> disposable) SubscribeAsync(
>    string topicName,
>    CancellationToken cancellationToken = default);```
Good luck awaiting that 😄 
Why return an `IDisposable` at all when you have an `IObservable` which returns `IDisposable` on `Subscribe`? This exact pattern came up earlier and we rejected it because it's confusing. I don't subscribe to a topic to get an observable that I can subscribe to. If you insist on observables then let's just add a method that can be used intuitively:
```cs
IObservable<RouterMessage> Topic(string topicName);
```
```
_subscription = messageRouter.Topic("symbol_selected")
    .Select(msg => SymbolSelectedMessage.FromRouterMessage(msg))
    .Where(msg => IsSymbolVisible(msg.Symbol))
    .Subscribe(msg => HandleSymbolSelectedMessage(msg));
```

I don't personally like this, and I'd rather have a messaging API that lets you process messages asynchronously by default, and build something around it if you want old school observables.

kruplm (approved):
LGTM - The moving and renaming of the web components and the demo applications can be part of an other PR since the original scope for this was to introduce the improvements on the ModuleLoader.

PR #96, opened 2022-09-14
created prototype with Avalonia docking panes hosting WebView2 browsers and wired the communications mechanism. 

nick-polyak-ms:


Tryouts/Plugins/Services/CommunicationsServices/GrpcCommunicationsService/SubscriptionClientImpl.cs
```diff
@@ -43,20 +42,45 @@ public async Task Publish<TMessage>(Topic topic, TMessage msg)
...
 
-        public async IAsyncEnumerable<TMessage> Subscribe<TMessage>(Topic topic)
-            where TMessage : IMessage, new()
+        private AsyncServerStreamingCall<ReturnedSubscriptionItem>? _replies;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="topic"></param>
+        /// <param name="messageType">Should have a default constructor and shoudl derive from IMessage!!!</param>
```
ztanczos, line 51: Nitpick: typo: shoudl

Tryouts/Plugins/Services/CommunicationsServices/GrpcCommunicationsService/SubscriptionClientImpl.cs
```diff
@@ -43,20 +42,45 @@ public async Task Publish<TMessage>(Topic topic, TMessage msg)
             await _client.PublishAsync(publishRequest);
         }
 
-        public async IAsyncEnumerable<TMessage> Subscribe<TMessage>(Topic topic)
-            where TMessage : IMessage, new()
+        private AsyncServerStreamingCall<ReturnedSubscriptionItem>? _replies;
```
ztanczos, line 45: Won't you need an instance for each topic? Can a client subscribe to one topic multiple times?

Tryouts/Prototypes/WV2Prototype/MainWindow.axaml
```diff
@@ -0,0 +1,55 @@
...
+						<ms:EmbeddedNativeControl ProcessExePath="Plugins/ApplicationPlugins/SimpleWpfApp/SimpleWpfApp.exe"/>
+					</np:DockItem>
+					<np:DockItem Header="Another Simple App">
+						<ms:EmbeddedNativeControl ProcessExePath="Plugins/ApplicationPlugins/AnotherWpfApp/AnotherWpfApp.exe"/>
+					</np:DockItem>
+				</np:StackDockGroup>
+				<np:StackDockGroup TheOrientation="Horizontal">
+					<np:DockItem Header="Web Publish Application">
+						<ms:EmbeddedWebView x:Name="PublishingWebView"
+											Source="file:///C:\MSDE\polyanic\git\ComposeUI_Web\Tryouts\Plugins\ApplicationPlugins\SimplePublishingWebApp\SimplePublishingWebApp.html"/>
```
ztanczos, line 45: This would only work on your machine

Tryouts/Prototypes/WV2Prototype/MainWindow.axaml
```diff
@@ -0,0 +1,55 @@
...
+					</np:DockItem>
+					<np:DockItem Header="Web Subscribe Application">
+						<ms:EmbeddedWebView x:Name="SubscribingWebView"
+											Source="file:///C:\MSDE\polyanic\git\ComposeUI_Web\Tryouts\Plugins\ApplicationPlugins\SimpleSubscribedWebApp\SimpleSubscribingWebApp.html"/>
```
ztanczos, line 49: Same here

Tryouts/Prototypes/WV2Prototype/MorganStanley.ComposeUI.Prototypes.WV2Prototype.csproj
```diff
@@ -0,0 +1,62 @@
...
+	</ItemGroup>
+	<ItemGroup>
+		<!--This helps with theme dll-s trimming.
+	If you will publish your application in self-contained mode with p:PublishTrimmed=true and it will use Fluent theme Default theme will be trimmed from the output and vice versa.
+	https://github.com/AvaloniaUI/Avalonia/issues/5593 -->
+		<TrimmableAssembly Include="Avalonia.Themes.Fluent" />
+		<TrimmableAssembly Include="Avalonia.Themes.Default" />
+	</ItemGroup>
+	<ItemGroup>
+		<!--Condition below is needed to remove Avalonia.Diagnostics package from build output in Release configuration.-->
```
ztanczos, line 48: As discussed earlier please remove these comments as they are no longer relevant.

Tryouts/Prototypes/WV2Prototype/MorganStanley.ComposeUI.Prototypes.WV2Prototype.csproj
```diff
@@ -0,0 +1,62 @@
...
+		<OutputPath>bin\Debug\net6.0</OutputPath>
+	</PropertyGroup>
+	<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+	  <Optimize>False</Optimize>
+	</PropertyGroup>
+	<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+	  <Optimize>True</Optimize>
+	</PropertyGroup>
+	<ItemGroup>
+	  <Compile Remove="EmbeddedWebView.cs" />
```
ztanczos, line 35: Can this be removed too?

Tryouts/Visuals/Windows/WindowsVisualUtils/EmbeddedWebView.cs
```diff
@@ -0,0 +1,166 @@
...
+            if ((!_domContentLoaded) || (!_objectsAdded))
+            {
+                return;
+            }
+
+            _webView.CoreWebView2.ExecuteScriptAsync("window.IsWebView2Ready = true;");
+            _webView.CoreWebView2.ExecuteScriptAsync("console.log('Initialized IsWebView2ready');");
+
+            string js =
+                @"if (window.OnWebView2Initialized){ window.OnWebView2Initialized(); }";
```
ztanczos, line 107: As far as I see this function is defined in one of the hosted web applications. The embedded web view control should not be aware of what is it hosting.

Tryouts/Visuals/Windows/WindowsVisualUtils/EmbeddedWebView.cs
```diff
@@ -0,0 +1,166 @@
...
+/// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and limitations under the License.
+/// 
+/// ********************************************************************************************************
+
+using Avalonia;
+using Avalonia.Controls;
+using Avalonia.Platform;
+using Microsoft.Web.WebView2.Core;
+using Microsoft.Web.WebView2.WinForms;
```
ztanczos, line 16: Why Windows Forms? Why not WPF?

Tryouts/Visuals/Windows/WindowsVisualUtils/JSCommunicationsClient.cs
```diff
@@ -0,0 +1,89 @@
...
+
+using Google.Protobuf;
+using Microsoft.Web.WebView2.WinForms;
+using MorganStanley.ComposeUI.Tryouts.Core.Abstractions;
+using Subscriptions;
+using System.Reactive.Linq;
+using System.Runtime.InteropServices;
+using System.Text.Json;
+
+namespace MorganStanley.ComposeUI.Prototypes.WV2Prototype
```
ztanczos, line 20: Inconsistent namespace - the file is in a different folder

Tryouts/Visuals/Windows/WindowsVisualUtils/EmbeddedWebView.cs
```diff
@@ -0,0 +1,166 @@
...
+
+namespace MorganStanley.ComposeUI.Tryouts.Visuals.Avalonia.WindowsVisualUtils
```
ztanczos, line 18: Inconsistent namespace: in the namespace you have Tryouts.Visuals.Avalonia. but the file is in Tryouts\Visuals\Windows\..

Tryouts/Visuals/Windows/WindowsVisualUtils/MorganStanley.ComposeUI.Tryouts.Visuals.Windows.WindowsVisualUtils.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
ztanczos, line 1: *Windows*VisualUtils - the *Windows* prefix is reduntant - it is already under .Visuals.*Windows*.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
```diff
@@ -18,10 +18,13 @@ public interface ISubscriptionClient
...
-        Task Publish<TMessage>(Topic topic, TMessage msg)
-             where TMessage : IMessage;
+        Task Publish(Topic topic, IMessage msg);
 
         IAsyncEnumerable<TMessage> Subscribe<TMessage>(Topic topic)
             where TMessage : IMessage, new();
+
+        IAsyncEnumerable<IMessage> Subscribe(Topic topic, Type messageType);
+
+        void Unsubscribe();
```
BalassaMarton, line 28: What is this method doing? Is it unsubscribing from everything? Who's going to call this method?

Tryouts/Core/Abstractions/ISubscriptionClient.cs
```diff
@@ -18,10 +18,13 @@ public interface ISubscriptionClient
     {
         Task Connect(string host, int port);
```
BalassaMarton, line 19: App devs should not be concerned about configuration values like host and port. The client shouldn't even have a `Connect` method. This should be set up once on startup (module load) and the client implementation should open the connection when necessary.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
```diff
@@ -18,10 +18,13 @@ public interface ISubscriptionClient
...
 
-        Task Publish<TMessage>(Topic topic, TMessage msg)
-             where TMessage : IMessage;
+        Task Publish(Topic topic, IMessage msg);
```
BalassaMarton, line 21: `IMessage` is a gRPC-specific type. This is against every requirement we set for the Message Router so far - messages should be plain text or raw bytes and the communication protocol should not leak into the API.
Topics are still defined as an enumeration in an abstractions package - application should be able to use arbitrary topics.

BalassaMarton (changes requested):

Tryouts/Visuals/Windows/WindowsVisualUtils/JSCommunicationsClient.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    [ClassInterface(ClassInterfaceType.AutoDual)]
+    [ComVisible(true)]
+    public class JSCommunicationsClient
```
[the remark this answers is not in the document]
ztanczos: This is a 'host object' for WebView2: https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addhostobjecttoscript?view=webview2-dotnet-1.0.1185.39

Tryouts/Visuals/Windows/WindowsVisualUtils/EmbeddedWebView.cs
nick-polyak-ms: I assume it will be a reserved name to use for doing browser initialization from C# code.

Tryouts/Visuals/Windows/WindowsVisualUtils/EmbeddedWebView.cs
nick-polyak-ms: WinForms controls have the Window Handle automatically - they do not require an extra step for obtaining the handle. It does not matter what we use for the WebView2 browser - the browser/shell communications are exactly the same.

Tryouts/Plugins/ApplicationPlugins/SimplePublishingWebApp/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "simple-web-app",
+  "version": "0.0.0",
+  "description": "SimpleWebApp",
+  "author": {
+    "name": ""
+  },
+  "dependencies": {
+    "jquery": "^3.6.1"
```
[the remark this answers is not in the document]
nick-polyak-ms: Do not think anything bad about good old JQuery especially for the sample as you stated.

Tryouts/Visuals/Windows/WindowsVisualUtils/JSCommunicationsClient.cs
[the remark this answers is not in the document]
nick-polyak-ms: It is Window Specific because currently WebView2 works only for Windows. They plan to make it multiplatform (and I hope they'll follow up on their promise) but not yet. 

AddHostObjectToScript method adds C# objects to become available from within the HTML scripts which I need in order to wire the communications.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: Eventually it should move over to a configuration module I agree. This is just a small prototype. I'll open another Jira to configure the end points.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: The client can call this method e.g. before the corresponding functionality is destroyed - e.g. when a docking pane is destroyed.

Tryouts/Plugins/Services/CommunicationsServices/GrpcCommunicationsService/SubscriptionClientImpl.cs
nick-polyak-ms: yes, client can subscribe for each topic multiple times

Tryouts/Prototypes/WV2Prototype/MainWindow.axaml
nick-polyak-ms: Yes, I will fix it tomorrow

Tryouts/Visuals/Windows/WindowsVisualUtils/JSCommunicationsClient.cs
nick-polyak-ms: fixed

Tryouts/Visuals/Windows/WindowsVisualUtils/EmbeddedWebView.cs
nick-polyak-ms: fixed

Tryouts/Visuals/Windows/WindowsVisualUtils/MorganStanley.ComposeUI.Tryouts.Visuals.Windows.WindowsVisualUtils.csproj
nick-polyak-ms: fixed

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: I changed the API, current Subscribe returns disposable unsubscriber. Also topics are now string.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: I completely changed ISubscriptionClient to be more in tune with Message router API:

    public interface ISubscriptionClient
    {
        Task Connect(string host, int port);

        Task Publish(string topic, string msg);

        (IObservable<CommunicationsMessage> messageStream, IDisposable unsubscriber)
            Subscribe(string topic);
    }

Published message is now string, topic is a string and Subscribe returns `CommunicationsMessage` that is exactly the same as `RoutedMessage` (once we integrate, I'll use `RoutedMessage`)
Please, see if this is ok with you.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
```diff
@@ -9,19 +9,15 @@
...
     {
         Task Connect(string host, int port);
 
-        Task Publish<TMessage>(Topic topic, TMessage msg)
-             where TMessage : IMessage;
+        Task Publish(string topic, string msg);
 
-        IAsyncEnumerable<TMessage> Subscribe<TMessage>(Topic topic)
-            where TMessage : IMessage, new();
+        (IObservable<CommunicationsMessage> messageStream, IDisposable unsubscriber)
```
ztanczos, line 20: This return type is really confusing. The caller would do
```
(var messageStream, var unsubscriber) = subscriptionClient.Subscribe("...");
```
and then 
```
var unsubscriber2 = messageStream.Subscribe();
```
so would end up with two `IDisposable` unsubscription tokens.

Tryouts/Visuals/Windows/VisualUtils/JSCommunicationsClient.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        private IDisposable? _disposable = null;
+        private IDisposable? _serverUnsubscriber = null;
+        private string _topic;
+
+        public void Subscribe(string topic)
+        {
+            _topic = topic;
+            _disposable?.Dispose();
```
ztanczos, line 53: This would mean that one client cannot subscribe to more than one topic - subsequent subscriptions would dispose the previous ones.

Tryouts/Plugins/ApplicationPlugins/SimpleNodeApp/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "simple-node-app",
+  "version": "0.0.0",
+  "description": "SimpleNodeApp",
+  "main": "app.js",
+  "author": {
```
kruplm, line 6: This node application can't run, because you have no execution step here. Right now it can only run in visual studio as a node.js project.

Tryouts/Plugins/ApplicationPlugins/SimpleSubscribedWebApp/SimpleSubscribingWebApp.html
```diff
@@ -0,0 +1,33 @@
...
+    <title>Simple Subscribing Web App</title>
+</head>
+<body>
+    <h1 id="MyHeader">Simple Subscribing Web App</h1>
+    <div id="TheText">blabla</div>
+    <script src="node_modules/jquery/dist/jquery.js" type="text/javascript"></script>
+    <script type="text/javascript">
+        window.OnWebView2Initialized = function () {
+            if (window.IsWebView2Ready && (!window.IsSubscribed)) {
+                var jsCommunicationsClient = window.chrome.webview.hostObjects.sync.JavaScriptCommunicationsClient;
```
kruplm, line 16: The web applications are communicating on the webview, by  accessing a lower level API of the browser. Which could be fine for simple demoing purposes.
However this POC does not prove that "web and .net applications can communicate via the same message router" (where the message router is powered by gRPC in this prototype)

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: I know it is a bit awkward and I'll try to improve it.

Tryouts/Plugins/ApplicationPlugins/SimpleNodeApp/package.json
nick-polyak-ms: this is essentially not a node application - it is started by the browser. It does not need to be build - only installed JQuery package. I am just not sure how to prevent it from building, but make it still install the package.

Tryouts/Plugins/ApplicationPlugins/SimpleSubscribedWebApp/SimpleSubscribingWebApp.html
nick-polyak-ms: Yes, Maria, you noticed correctly, I limited the RPC communications to between two C# programs, but the wiring of C# to WebView2 is pretty generic and I can prove that communication by any messages is possible and even more so if we limit the messages to Topic and Json-string tuple. 

In particular, try to come up with a some complex communication and I'll show how to achieve that.

Tryouts/Visuals/Windows/WindowsVisualUtils/EmbeddedWebView.cs
bingenito: It could matter, and perhaps also be beneficial to use winforms.  With cefsharp we found with winforms we could have 60fps vs 30fps of wpf hosting and there were also less issues with keyboard focus, etc.

Tryouts/Plugins/ApplicationPlugins/SimpleNodeApp/package.json
kruplm: You aren't creating a bundle from this package and - unless I haven't spotted it - nothing serves this demo application.
So essentially an html file gets loaded to the webview over the file:// protocol.

Tryouts/Plugins/ApplicationPlugins/SimplePublishingWebApp/package.json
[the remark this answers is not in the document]
ZKRobi: As I said, I don't mind, I was just expressing surprise :)

Tryouts/Visuals/Windows/WindowsVisualUtils/JSCommunicationsClient.cs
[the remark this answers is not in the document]
ZKRobi: Got it, thanks for the explanation

Tryouts/Plugins/ApplicationPlugins/SimpleSubscribedWebApp/SimpleSubscribingWebApp.html
ZKRobi: I'm concerned that this way the web app depends on being displayed in a WebView2 browser.
If I think about a web app written for compose, one of the biggest benefits I'd assume is that it's the most shell-independent technology I can think of, as long as the shell supports some kind of web browser, while other technologies might not always work that "natively".
By coding the communication in c# and tying it to the webview api, this benefit seems to be completely lost.

Tryouts/Plugins/ApplicationPlugins/SimpleNodeApp/package.json
nick-polyak-ms: yes, exactly - but there is a twist to it - it needs to unpack the node JQuery package.

Tryouts/Visuals/Windows/VisualUtils/JSCommunicationsClient.cs
nick-polyak-ms: I'll take a look at expanding this functionality to cover multiple topics for a single client.

Tryouts/Prototypes/WV2Prototype/MainWindow.axaml
nick-polyak-ms: fixed

Tryouts/Prototypes/WV2Prototype/MorganStanley.ComposeUI.Prototypes.WV2Prototype.csproj
nick-polyak-ms: removed

Tryouts/Prototypes/WV2Prototype/MorganStanley.ComposeUI.Prototypes.WV2Prototype.csproj
nick-polyak-ms: fixed

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: I simplified the API further to 

     IObservable<CommunicationMessage> Subscribe(string topic)

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: Here is a new github issue covering this: https://github.com/morganstanley/ComposeUI/issues/98

Tryouts/Visuals/Windows/VisualUtils/JSCommunicationsClient.cs
nick-polyak-ms: The functionality now can subscribe and unsubscribe from multiple topics

        public void Subscribe(string topic)
        {
            if (_subscriptions.ContainsKey(topic))
            {
                return; // subscriptions already exists
            }

            var observable =
                _subscriptionClient
                    .Subscribe(topic);

            IDisposable disposable = observable.Subscribe(OnMessageArrived);

            _subscriptions.Add(topic, disposable);
        }


        public void Unsubscribe(string topic)
        {
            if (_subscriptions.ContainsKey(topic))
            {
                IDisposable disposable = _subscriptions[topic];

                disposable.Dispose();

                _subscriptions.Remove(topic);
            }
        }

Tryouts/Core/Abstractions/ISubscriptionClient.cs
```diff
@@ -1,27 +1,11 @@
-﻿/// ********************************************************************************************************
```
ztanczos, line 1: Nitpick: please don't remove the license header

Tryouts/Core/Abstractions/ISubscriptionClient.cs
```diff
@@ -1,27 +1,11 @@
...
     {
         Task Connect(string host, int port);
 
-        Task Publish<TMessage>(Topic topic, TMessage msg)
-             where TMessage : IMessage;
+        Task Publish(string topic, string msg);
 
-        IAsyncEnumerable<TMessage> Subscribe<TMessage>(Topic topic)
-            where TMessage : IMessage, new();
+        IObservable<CommunicationsMessage> Subscribe(string topic);
```
ztanczos, line 9: I'm not sure 'Subscribe' would be the best name for this method since it wouldn't actually subscribe to anything - you'd need to call Subscribe() on the returned IObservable. @BalassaMarton suggested that this method could be called 'Topic' or something similar - that would make the code more readable.

On the other hand I'm a bit ambivalent about IObservable vs IAsyncEnumerable. I understand that one can be converted to the other and vice-versa, but looking at the current implementation of this interface it'd feel more _natural_ to simply return IAsyncEnumerable instead of IObservable.

Tryouts/Plugins/ApplicationPlugins/SimpleSubscribedWebApp/SimpleSubscribingWebApp.html
ztanczos: What are the advantages of using gRPC with this approach if the communication interface is via WebView2 interop objects and json serialization?

Tryouts/Visuals/Windows/VisualUtils/JSCommunicationsClient.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using Google.Protobuf;
+using Microsoft.Web.WebView2.WinForms;
+using MorganStanley.ComposeUI.Tryouts.Core.Abstractions;
+using Subscriptions;
+using System.Reactive.Linq;
+using System.Runtime.InteropServices;
+using System.Text.Json;
+
+namespace MorganStanley.ComposeUI.Tryouts.Visuals.Windows.VisualUtils
```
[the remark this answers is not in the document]
nick-polyak-ms: It depends on EmbeddedWebView - therefore it depends on both Windows and Avalonia. It does not make sense to create another project for just one class. So I think Windows.VisualUtils is the best place for it, at least for now.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: I think we should go by what the clients need, rather than by implementation. Observable is smth people are comfortable - it has been around for longer and there are plenty examples. IAsyncEnumerable - I am not even sure if it is possible to use Linq operators on them without converting them first to IObservable. So I think in order to make it easier for the clients - it is better to keep them as IObservable.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: restored

Tryouts/Core/Abstractions/ISubscriptionClient.cs
ZKRobi: This still results in awkward code like MessageRouter.Subscribe("myTopic").Subscribe(observer);

Tryouts/Core/Abstractions/ISubscriptionClient.cs
ZKRobi: I agree renaming this method to Topic, or something similar, currently this results in MessageRouter.Subscribe("topic").Subscrive(observer);

For me, IObservable feels more natural, but I guess it's a matter of taste and the use-case, I don't mind either too much.

Tryouts/Visuals/Windows/VisualUtils/JSCommunicationsClient.cs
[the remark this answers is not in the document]
ztanczos: In its current form yes it depends on WebView2 but it doesn't need to. It could instead be used by a visual component.

Tryouts/Core/Abstractions/ISubscriptionClient.cs
nick-polyak-ms: And I changed the method name to ConsumeTopic(string topic)

Tryouts/Plugins/ApplicationPlugins/SimpleSubscribedWebApp/SimpleSubscribingWebApp.html
nick-polyak-ms: I wanted to minimize the load on WebView2, because it is kind of messy. Also, it only works for the browser projects, not for  WPF or Forms or Avalonia.

nick-polyak-ms: Here is an expanded answer:
I think I understand your question - WebView2 object  is not a singleton. There is one WebView2 object per browser, so if you e.g. trigger a PostMessage event in one WebView2 object - the other will not react. Of course, they run in the same process and potentially can be connected without GRPC but that would be extra software and extra work to connect them (which we can anyways do in the future if needed).

ztanczos: I see. The _other_ MessageRouter works in a way that web clients can directly connect to the WebSocket endpoint and the communication flows like this: web/.NET <--> MessageRouter <--> web/.NET. 
With this approach the communication flows: web <--> WebView2 <--> gRPC messaging client <--> gRPC messaging service <--> WebView2 <--> web.

I think this brings up a good question that we should discuss: do we want to restrict/recommend hosted web clients to connect directly to the MessageRouter or should/could it also go through WebView2 interop objects?

nick-polyak-ms: I think that uniformity has advantage by having fewer 'if' branches and by increasing the reuse and making the code smaller and easier to maintain, debug and expand. I this case, a browser application (with a small WebView2 adapter) communicates in very much the same way as Forms, WPF and Avalonia applications. 

It does not preclude the browser application using WebSockets or some other means of communications for its own purpose.

ztanczos: I think uniformity is an advantage that exists in both cases, I don't see a difference w.r.t. uniformity as both have well defined API surface. Can you please elaborate why do you think an extra WebView2 layer provides more reuse or makes the code smaller?

npolyak: I meant uniformity between Desktop App (WPF/Forms/Avalonia) and Browser case. For example, we need to add a message that would work between any two apps (say link message for linking two applications so that the output of one will be the input for the other) - we'll be able to add it simply by modifying our communications layer and it will work for WPF the same as it works for browser. Now if we have a separate path for a browser, we'll have to modify it in two places, right? The same is true about debugging.

PR #97, opened 2022-09-19
Glossary initial commit

ZKRobi:


BalassaMarton (approved):

ztanczos (approved):

psmulovics (commented):
Do we want to move glossary under docs/ or architecture/?

kruplm:
> Do we want to move glossary under docs/ or architecture/?

No need. Most OSS projects have the glossary.md (or a folder) in the root, so it's easy to find.

ztanczos:
> Do we want to move glossary under docs/ or architecture/?

I wouldn't move it under architecture. If / once we'd have significant amount of docs then we can re-visit whether to put glossary there.

Glossary/Plugin.md
```diff
@@ -0,0 +1,5 @@
+#Plugin
```
BalassaMarton, line 1: Spaces are missing after `#` (it is needed for correct rendering)

BalassaMarton:
Minor thing: can we use capitalization consistently? Most of the root directories are all lower case, with some irritating exceptions.

ZKRobi:
> 

Sounds good, but which one to use? :)

kruplm (approved):
LGTM.

Glossary/Plugin.md
ZKRobi: fixed

BalassaMarton (approved):

glossary/README.md
```diff
@@ -0,0 +1,7 @@
+# Glossary
+This folder contains terms used in the Compose project to describe certain concepts.
+
+[Application](Application.md)\
```
BalassaMarton, line 4: I say we drop the links because we will forget to update the list for sure. The link to Manifest is missing already.

ztanczos: Can this be auto-generated somehow?

ZKRobi: I'm dropping the links for now, and I'll look into a solution for generating them. I could only find a solution for generating a table of contents within a single file when I did a quick search

kruplm (approved):

PR #107, opened 2022-11-24
MessageRouter MVP

BalassaMarton:
- Basic access token handling
- Refactoring


BalassaMarton:
I recommend reviewing this commit by commit as it contains a massive namespace change that doesn't require review

Tryouts/Messaging/Client/MessageRouterClient.cs
```diff
@@ -70,10 +82,11 @@ public ValueTask ConnectAsync(CancellationToken cancellationToken = default)
         string? payload = null,
         CancellationToken cancellationToken = default)
     {
-        var requestId = Guid.NewGuid().ToString();
+        var requestId = new string(Guid.NewGuid().ToString());
```
kruplm, line 85: Please double check if new string allocation is really needed here.

BalassaMarton: That's definitely not right, I'll fix it.

kruplm (dismissed):

kruplm (approved):

kruplm (approved):

kruplm (approved):

PR #112, opened 2022-12-13
MessageRouter MVP - Scopes and Namespace cleanup

BalassaMarton:
This PR contains 2 commits. 

First commit is the introduction of the `MessagingScope` type which can be passed to `Publish` but is ignored at the moment.

Second commit is a massive namespace change where only names and namespaces were changed, and some types were moved from the Client package to the Core package.


Tryouts/Messaging/Client/PublishOptions.cs
```diff
@@ -10,10 +10,11 @@
...
+using MorganStanley.ComposeUI.Messaging.Core;
+
 namespace MorganStanley.ComposeUI.Messaging.Client;
 
-// TODO: MessagingScope
-public class MessagingScope
+public struct PublishOptions
 {
-    
+    public MessagingScope Scope { get; set; }
```
ztanczos: Mutable structs aren't a good idea

BalassaMarton: I have no problem changing it to init only, but since the options will be extended over time, the alternative is a positional record with many optional parameters. It can also be a class, if we don't mind the extra allocation per message just for a parameter object.

ztanczos: init only sounds like a much safer choice to me.

kruplm (dismissed):

kruplm (approved):

PR #132, opened 2023-01-25
Message Router JS client

BalassaMarton:
Initial version of the MessageRouter client with WebSocket binding
Library structure and implementation closely follows the .NET client.


Tryouts/Messaging-JS/src/client/MessageRouterClient.ts
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+
+    next(message: TopicMessage): void {
+        for (let subscriber of this.subscribers) {
+            try {
+                subscriber.next?.call(subscriber, message);
+            }
+            catch (err) {
+                console.error(err);
```
ztanczos, line 420: I don't know what's the preferred way for logging errors - is console.error() a safe default choice?

kruplm: Yes it is. If you log with the appropriate intent you can filter by log levels on the developer console, etc.

Tryouts/Messaging-JS/src/InvokeOptions.ts
```diff
@@ -0,0 +1,6 @@
+import { MessageScope } from ".";
```
kruplm, line 1: Please Don't import modules like` from "."` and ` from ".."` (this applies for every occurrence, but I won't repeat myself everywhere)

Tryouts/Messaging-JS/src/client/MessageRouterClient.spec.ts
```diff
@@ -0,0 +1,597 @@
+import { MessageRouterClient, ClientState, Connection, OnMessageCallback, OnErrorCallback, OnCloseCallback } from ".";
+import { MessageContext, MessageHandler, MessageScope, TopicMessage, TopicSubscriber } from "..";
+import * as messages from "../protocol/messages";
+import "jest-extended"
+import { DuplicateEndpointError, ErrorTypes, MessageRouterError } from "../exceptions";
+import { Error } from "../protocol";
+
+describe("MessageRouterClient", () => {
```
kruplm, line 8: Nitpick but you have a lot of empty new lines in tests.

kruplm (changes requested):

Tryouts/Messaging-JS/src/InvokeOptions.ts
BalassaMarton: Yeah I wasn't sure about that. Should I really import everything from its own file? Does it still make sense to have and index.ts which re-exports everything? I did that so that consuming packages can import everything from a few modules.

Tryouts/Messaging-JS/src/client/MessageRouterClient.spec.ts
BalassaMarton: That's intentional to improve readability.

kruplm (approved):

PR #134, opened 2023-01-31
Node launcher

kruplm:
Story: "As a frontend developer I can configure the height, width, title and the URL of a browser window from my JavaScript application."

- the shell can be launched from Node.JS as well.
- width, height and title of the window can be specified.
- An url can be passed via configuration or by the LoadUrl() function
- The default config loading (with pre-existing manifest file) still works the same way - so that our future Universal hosting goals can be achieved.
- As part of this PR the manifest file parsing was also improved.


Tryouts/Prototypes/Shell/Node-launcher/Lib/src/Launcher.ts
```diff
@@ -0,0 +1,45 @@
...
+                this.argsArray.push(`--height=${_config?.height}`);
+            }
+        }
+
+        console.debug("\t[]", this.argsArray);
+    }
+
+    
+    public launch() {
+        //TODO replace after application is properly packaged and added to PATH
```
BalassaMarton, line 35: Can we remove these early on? Just add the directory to your PATH variable (and document it).

Tryouts/Prototypes/Shell/App.xaml.cs
```diff
@@ -28,6 +28,11 @@ namespace Shell
     /// </summary>
     public partial class App : Application
     {
-        
+        public string[] CustomerCode { get; set; }
```
BalassaMarton, line 31: What customer code? Who's the customer and what kind of code is provided here? It's actually populated from command line arguments, so why isn't it called `CommandLineArguments`?

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
```diff
@@ -40,13 +41,59 @@ public partial class MainWindow : Window
...
+            if (commands.ContainsKey("url"))
+            {
+                this.commandLineURL = commands["url"];
+            }
+            else 
+            {
+                this.commandLineURL = "about:blank";
+            }
+
+            var lazyWebContent = new WebContent(this.commandLineURL);
```
BalassaMarton, line 87: Don't call it lazy if it's not a `Lazy<T>` or similar.

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
```diff
@@ -40,13 +41,59 @@ public partial class MainWindow : Window
...
+        }
+
+        private void LazyLoading() {
+            Dictionary<string, string> commands = new Dictionary<string, string>();
+
+            this.hereIsMyCode.ToList<string>().ForEach(item => {
+                item = item.TrimStart('-');
+                string[] command = item.Split("=");
+
+                commands.Add(command[0], command[1]);
```
BalassaMarton, line 75: This will add the wrong url if it contains the '=' character, which is common.

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
```diff
@@ -40,13 +41,59 @@ public partial class MainWindow : Window
...
+            });
```
BalassaMarton, line 76: Don't invent the wheel, use the `System.CommandLine` library instead, it's well tested, robust, and pretty standard.

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
```diff
@@ -40,13 +41,59 @@ public partial class MainWindow : Window
...
+
+            if (commands.ContainsKey("url"))
```
BalassaMarton, line 78:
```cs
if (commands.TryGetValue("url", out var url) 
{
    this.commandLineURL = url;
}
```
or
```
this.commandLineURL = commands.GetValueOrDefault("url");
```
etc. 
But you should just ditch the whole thing and use `System.CommandLine` or similar parsing library.

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
```diff
@@ -40,13 +41,59 @@ public partial class MainWindow : Window
         private ManifestModel config;
         private ModuleModel[]? modules;
 
+        private string commandLineURL;
+        private string[] hereIsMyCode = (Application.Current as App).CustomerCode;
```
BalassaMarton, line 45: Don't invent names. Rename to `commandLineArgs` or something that says what it does.

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
```diff
@@ -40,13 +41,59 @@ public partial class MainWindow : Window
...
 
-            config = new ManifestParser().Manifest;
-            modules = config.Modules;
-            DataContext = modules;
+            this.Loaded += MainWindow_Loaded;
+        }
+
+        private void MainWindow_Loaded(object sender, RoutedEventArgs e)
+        {
+            if (this.hereIsMyCode.Length != 0)
```
BalassaMarton, line 56: What's the expected behavior here? If I provide the url, an empty `MainWindow` will pop up. Consider hiding the main window in this case.

Tryouts/Prototypes/Shell/Node-launcher/Lib/rollup.config.js
```diff
@@ -0,0 +1,13 @@
+﻿import { nodeResolve } from '@rollup/plugin-node-resolve';
+import typescript from '@rollup/plugin-typescript';
+
+export default {
+    input: 'src/index.ts',
+    output: {
+      
+        file: 'output/bundle.mjs',
+        format: 'es',
+        name: "@compose/node-launcher"
```
BalassaMarton, line 10: Why is the scope and name different from the one in `package.json`? Is this something rollup-specific?

Tryouts/Prototypes/Shell/Node-launcher/Lib/src/BrowserWindow.ts
```diff
@@ -0,0 +1,25 @@
+export class BrowserWindow {
```
BalassaMarton, line 1: What's this class for? Is it a stub for future implementation?

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
ztanczos, line 45: If you know the type of Application.Current is always App then use direct cast. If you don't know/can't guarantee that then using `as` is correct, however you should check for null result then.

Tryouts/Prototypes/Shell/Manifest/ManifestParser.cs
```diff
@@ -29,7 +30,13 @@ internal class ManifestParser
...
+            string fileName = "exampleManifest.json";
+            string processPath = Environment.ProcessPath;
+            string folder= Path.GetDirectoryName(processPath);
+
+            string path = Path.Combine(folder, @"Manifest\", fileName);
+            
+            OpenManifestFile(path);
         }
 
         public async void OpenManifestFile(string manifestFile)
```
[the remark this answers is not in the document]
ztanczos: +1, but also don't use `async void` methods, except for evend handlers. See https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming for details.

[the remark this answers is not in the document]
BalassaMarton: In defense of `async void`, if it's clear that it won't be awaited and you handle exceptions inside the method, it should be fine.

Tryouts/Prototypes/Shell/Node-launcher/Lib/src/BrowserWindow.ts
kruplm: Yes, I accidentally committed that WIP code after I rearranged the project files for the monorepo.
I've pushed the intended code just now.

Tryouts/Prototypes/Shell/Node-launcher/Lib/rollup.config.js
kruplm: It's the global variable name that represents the bundle if it's iife or umd, But it's not needed for ES so I'm going to remove it.

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
kruplm: First time I checked System.CommandLine, it was still in preview. Other parsing libraries have licenses that aren't good for this project, or hasn't been maintained for a while.
My idea was to switch to System.CommandLine, and rather deal with it's changes in the future. But I wanted to discuss it with the team first.

ztanczos: We should be good to use System.CommandLine, that package is already being used by large projects (e.g.: dotnet CLI itself).

kruplm: That sounds reassuring.

Tryouts/Prototypes/Shell/Manifest/ManifestParser.cs
```diff
@@ -20,26 +20,28 @@
...
 using System.Windows.Controls.Primitives;
+using System.Windows.Documents;
 
 namespace Manifest
 {
-    internal class ManifestParser
+    internal static class  ManifestParser
     {
-        internal ManifestModel Manifest { get; set; }
+        internal static ManifestModel Manifest { get; set; }
```
BalassaMarton, line 29: Unnecessary property. Why would you have a static parser that also stores the last manifest it has parsed?

Tryouts/Prototypes/Shell/Manifest/ManifestParser.cs
```diff
@@ -20,26 +20,28 @@
...
 
-        public ManifestParser()
+        public static  ManifestModel OpenManifestFile(string fileName)
```
BalassaMarton, line 31: `OpenManifestFile` suggests it returns something that can be later closed. Just call it `Parse` or `ParseManifest`.

Tryouts/Prototypes/Shell/Manifest/ManifestParser.cs
```diff
@@ -20,26 +20,28 @@
...
-        }
+            string processPath = Environment.ProcessPath;
+            string folder = Path.GetDirectoryName(processPath);
+            string path = Path.Combine(folder, @"Manifest\", fileName);
 
-        public async void OpenManifestFile(string manifestFile)
-        {
-             using (FileStream stream = File.Open(manifestFile, FileMode.Open))
-             {
+            using (FileStream stream = File.Open(path, FileMode.Open))
```
BalassaMarton, line 37: Suggestion:
```cs
using (var stream = File.OpenRead(path))
```
Edge case, but the original version will fail if you start two instances of the program quickly, because it tries to open the file for read/write access and without sharing.

Tryouts/Prototypes/Shell/Node-launcher/Lib/src/Launcher.ts
```diff
@@ -0,0 +1,39 @@
...
+import { WindowConfig } from './WindowConfig';
+
+export class Launcher {
+    private argsArray: string[] = [];
+
+    constructor(_config?: WindowConfig ) { 
+
+    }
+
+    private processArgs(_config?: WindowConfig) {
```
BalassaMarton, line 11: I'm confused about naming conventions here. Why the underscore before the function parameter?

Tryouts/Prototypes/Shell/Node-launcher/Lib/src/WindowConfig.ts
```diff
@@ -0,0 +1,5 @@
+export interface WindowConfig {
+    url: string;
```
BalassaMarton, line 2: Is this config object all or nothing? All these properties seem to be optional (and they are all treated as optional).

Tryouts/Prototypes/Shell/Properties/launchSettings.json
```diff
@@ -0,0 +1,7 @@
+{
```
BalassaMarton, line 1: Is this file really necessary?

BalassaMarton (changes requested):

Tryouts/Prototypes/Shell/MainWebWindow.xaml.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Windows.Shapes;
+
+namespace Shell
+{
+    /// <summary>
+    /// Interaction logic for MainWebWindow.xaml
+    /// </summary>
+    public partial class MainWebWindow : Window
+    {
+        private string[] commandLineArguments = (Application.Current as App).CommandLineArguments;
```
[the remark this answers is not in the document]
ztanczos: Another alternative: command line arguments could be passed to the objects which need them via the ctor.

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
```diff
@@ -71,7 +72,7 @@ private void ShowChild_Click(object sender, RoutedEventArgs e)
         {
             var context = (sender as Button).DataContext;
```
ztanczos, line 73: In this case the sender's `DataContext` is the same as the current Window's `DataContext`. But an even better approach would be to store the DataContext in a typed private field so that you wouldn't need to cast.

Tryouts/Prototypes/Shell/App.xaml
```diff
@@ -8,12 +8,13 @@ is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND
 See the License for the specific language governing permissions and limitations under the License.
 ********************************************************************************************************
 -->
-
+<!--StartupUri="MainWindow.xaml"-->
```
ztanczos, line 11: Please remove commented out code

Tryouts/Prototypes/Shell/Node-launcher/Lib/src/WindowConfig.ts
kruplm: The parameters are optional because the user might want to set the url, but not the width/height. And vice versa.

Tryouts/Prototypes/Shell/App.xaml.cs
```diff
@@ -28,6 +28,23 @@ namespace Shell
     /// </summary>
     public partial class App : Application
     {
-        
+        public string[] CommandLineArguments { get; set; }
```
ztanczos, line 31: This property is not used anymore, you can remove it.

Tryouts/Prototypes/Shell/MainWebWindow.xaml.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Security.Policy;
+
+namespace Shell
+{
+    /// <summary>
+    /// Interaction logic for MainWebWindow.xaml
+    /// </summary>
+    public partial class MainWebWindow : Window
+    {
+        public string Url;
```
ztanczos, line 26: Unused public field - you can remove it.

Tryouts/Prototypes/Shell/CommandLineArgumentsParser.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public CommandLineArgumentsParser(string[] _commandLineArguments)
+        {
+            commandLineArguments = _commandLineArguments;  
+        }
+        public async void CreateDynamicWindowWithOptions()
+        {
+            var titleOption = new Option<string>("--title", description: "Set title for window");
+            var urlOption = new Option<string>("--url", description: "Set url for webview. default: about:blank");
+            var widthOption = new Option<int>("--width", description: "Set width for window");
```
ztanczos, line 23: Do we have any defaults for width/height?

Tryouts/Prototypes/Shell/CommandLineArgumentsParser.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    internal class CommandLineArgumentsParser
+    {
+        private string[] commandLineArguments;
+
+        public CommandLineArgumentsParser(string[] _commandLineArguments)
+        {
+            commandLineArguments = _commandLineArguments;  
+        }
+        public async void CreateDynamicWindowWithOptions()
```
ztanczos, line 19: I'd change this to return a Task instead of async void to improve testability.

Tryouts/Prototypes/Shell/CommandLineArgumentsParser.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            var titleOption = new Option<string>("--title", description: "Set title for window");
+            var urlOption = new Option<string>("--url", description: "Set url for webview. default: about:blank");
+            var widthOption = new Option<int>("--width", description: "Set width for window");
+            var heightOption = new Option<int>("--height", description: "Set height for window");
+
+            var rootCommand = new RootCommand();
+
+            rootCommand.Add(titleOption);
```
ztanczos, line 28: Nitpick: you can use the collection initializer syntax here:
```
var rootCommand = new RootCommand
{
  urlOption,
  widthOption,
...
};
```

Tryouts/Prototypes/Shell/MainWebWindowOptions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shell
+{
+    public sealed class MainWebWindowOptions
+    {
+        public string Title { get; set; }
```
BalassaMarton, line 11: If any of these properties are optional, they should have nullable type like `string?` or `int?`

Tryouts/Prototypes/Shell/MainWebWindowOptionsParser.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public MainWebWindowOptions Parse(string[] args)
+        {
+            Parser parser = new Parser(rootCommand);
+            ParseResult parseResult = parser.Parse(args);
+
+            MainWebWindowOptions options = new MainWebWindowOptions
+            {
+                Title = parseResult.GetValueForOption(titleOption) ?? "Compose Web Container",
```
BalassaMarton, line 36: Some tests would be useful, I'm not sure what value is returned when the option is not set (it might even be an exception that we probably don't want)

Tryouts/Prototypes/Shell/MainWebWindowOptionsParser.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.CommandLine.Parsing;
+using System.Linq;
+using System.Security.Policy;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Media.Media3D;
+
+namespace Shell
+{
+    internal class MainWebWindowOptionsParser
```
BalassaMarton, line 14: This class should be static.

Tryouts/Prototypes/Shell/CommandLineArgumentsParser.cs
kruplm: The nitpick helped to turn the class to static, I needed to have a constructor before for this.

Tryouts/Prototypes/Shell/MainWebWindowOptions.cs
kruplm: since Window.Width and Height are double (fixed the wrong type now) instead of double? I can't set it double? here because I'd need to cast where I assign it.  These are never null because the default value is 0.0. 
And if the user doesn't specify the value from the JavaScript API the window is set to 450x800 now.

BalassaMarton (approved):

Tryouts/Prototypes/Shell/MainWebWindowOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Shell
+{
+    public sealed class MainWebWindowOptions
+    {
+        public double? Height { get; set; }
+        public string? Title { get; set; }
+        public string? Url { get; set; }
+        public double? Width { get; set; }
+
+        public double DefaultHeight { get; set; } = 450d;
```
BalassaMarton, line 16: These should be static constants with non-nullable type (`public const double DefaultHeight = 450` etc.)

BalassaMarton (changes requested):

BalassaMarton (approved):

PR #142, opened 2023-02-06
Setting up node build for Lerna monorepo

kruplm:


BalassaMarton:
Any reason why we have a separate workflow for tryouts?

kruplm:
> Any reason why we have a separate workflow for tryouts?

We have two workflows because of a previous agreement, when the current structure was created.
When we create the final structure for the repository the workflow can change.
It still can be a good idea to have something separate for our canaries in the future, but it can be done in a different way.

BalassaMarton (approved):

BalassaMarton (approved):

PR #143, opened 2023-02-06
Set up Lerna for Web Modules

kruplm:


ZKRobi (approved):

BalassaMarton:
Do we need this to be in the root of Tryouts (and later in src root)?

kruplm:
> Do we need this to be in the root of Tryouts (and later in src root)?

yes we do.

PR #148, opened 2023-02-08
Message Router - new error semantics

BalassaMarton:
Got rid of separate exception class per error type.
Set up yarn, jest and tsc to work properly with vs code.


kruplm:
The TypeScript code looks good to me

lilla28 (approved):

kruplm (approved):

PR #150, opened 2023-02-14
Message Router - better error handling, handle disconnected clients

BalassaMarton:
More consistent handling of errors when the connection drops from either side.
To validate, run the modules prototype and start exiting/killing modules.
Errors should be logged on the server only when the client is killed without properly closing the connection. The server should not crash in either case.

### DIFF
--- a/Tryouts/Messaging/Client/Client/WebSocket/WebSocketConnection.cs
+++ b/Tryouts/Messaging/Client/Client/WebSocket/WebSocketConnection.cs
@@ -20,6 +20,7 @@ using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Microsoft.IO;
 using MorganStanley.ComposeUI.Messaging.Client.Abstractions;
+using MorganStanley.ComposeUI.Messaging.Exceptions;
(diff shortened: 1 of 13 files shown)


BalassaMarton:
Putting this into draft status until the synchronization issues are solved (most probably by naively locking on everything)

Tryouts/Messaging/Client/Client/WebSocket/WebSocketConnection.cs
```diff
@@ -50,23 +51,65 @@ public async ValueTask DisposeAsync()
...
-        await _webSocket.ConnectAsync(_options.Value.Uri, cancellationToken);
-        StartReceivingMessages();
-        StartSendingMessages();
+        try
+        {
+            _webSocket = new ClientWebSocket();
+            await _webSocket.ConnectAsync(_options.Value.Uri, cancellationToken);
+            _ = Task.Run(ReceiveMessages);
+        }
+        catch (Exception e)
```
ztanczos, line 60: Nitpick: unused variable `e`

ztanczos (approved):

ztanczos (approved):

PR #160, opened 2023-02-22
Add Ribbon menu

kruplm:
* MainWindow has a dynamic ribbon menu


Tryouts/Prototypes/Shell/MainWindow.xaml
```diff
@@ -8,32 +8,43 @@ is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND
...
-            </ItemsControl.ItemTemplate>
-        </ItemsControl>
-        <StackPanel DockPanel.Dock="Top" Orientation="Horizontal" HorizontalAlignment="Center"></StackPanel>
-    </DockPanel>
-</Window> 
+        Title="Compose" Height="450" Width="600">
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition/>
```
ztanczos, line 22: You are only using one Row (row no. 0), the other one is unused. Why do you need two rows?

Tryouts/Prototypes/Shell/MainWindow.xaml
```diff
@@ -8,32 +8,43 @@ is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND
...
+        </Grid.RowDefinitions>
+        <Ribbon x:Name="RibbonWin"  SelectedIndex="0" Height="auto" >
```
ztanczos, line 24: Nitpick: "Auto" - although it is case-insensitive, the convention is to use CamelCased values

ztanczos (approved):

Tryouts/Prototypes/Shell/MainWindow.xaml
kruplm: Changed to CamelCase

psmulovics (dismissed):

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
```diff
@@ -12,18 +12,32 @@
...
+using System.Windows.Controls.Ribbon;
+using System.Windows.Controls.Ribbon.Primitives;
 
 namespace Shell
 {
     /// <summary>
     /// Interaction logic for MainWindow.xaml
     /// </summary>
-    public partial class MainWindow : Window
+    public partial class MainWindow : RibbonWindow
```
ZKRobi, line 40: This is not a critique for this PR, or the current state and goals, just a note that in the future, we probably should devise a more flexible solution for adding visual features than inheritance.

Tryouts/Prototypes/Shell/MainWindow.xaml
```diff
@@ -8,32 +8,42 @@ is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND
...
-        </ItemsControl>
-        <StackPanel DockPanel.Dock="Top" Orientation="Horizontal" HorizontalAlignment="Center"></StackPanel>
-    </DockPanel>
-</Window> 
+        Title="Compose" Height="450" Width="600">
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+        </Grid.RowDefinitions>
+        <Ribbon x:Name="RibbonWin"  SelectedIndex="0" Height="Auto" >
```
BalassaMarton, line 23: Suggestion: just call it `Ribbon`. No need for inventing names that don't reflect the component's purpose. It's not a window, so why would you append `Win` to its name, especially if the parent IS a `RibbonWindow`.

Tryouts/Prototypes/Shell/MainWindow.xaml.cs
BalassaMarton: This is fine for short term demoing purposes, but we will probably not have a ribbon control on the main window by default. As discussed earlier (and I mean years ago), we should provide a very minimalistic shell at first, concentrating on docking and layout persistence, and maybe later we can start adding features like a ribbon chrome.

Tryouts/Prototypes/Shell/MainWindow.xaml
```diff
@@ -8,32 +8,42 @@ is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND
...
 -->
-
-<Window x:Class="Shell.MainWindow"
+<RibbonWindow x:Class="Shell.MainWindow"
         xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-        xmlns:local="clr-namespace:Shell"
+        xmlns:local="clr-namespace:Shell" xmlns:primitives="clr-namespace:System.Windows.Controls.Ribbon.Primitives;assembly=System.Windows.Controls.Ribbon"
```
ZKRobi, line 16: This include seems unnecessary

BalassaMarton (approved):

Tryouts/Prototypes/Shell/MainWindow.xaml
ztanczos: Argh. My bad. But you get what I meant :)

BalassaMarton (approved):

ztanczos (approved):

PR #161, opened 2023-02-22
Shell + Message Router integration

BalassaMarton:
Message Router can now be created without any configuration in JS, just call `createMessageRouter()`.
Fixed async start and stop logic in Shell


Tryouts/Messaging-JS/.vscode/templates/js.lict
```diff
@@ -0,0 +1,9 @@
+Morgan Stanley makes this available to you under the Apache License,
```
psmulovics, line 1: Should we add lict files for the .NET part too?

BalassaMarton: We can, but I've been adding the license headers with ReSharper, and they are not compatible. I'm not sure if the vscode extension I used here is something widely used or just a random first hit.

Tryouts/Prototypes/Shell/App.xaml.cs
```diff
@@ -12,77 +12,160 @@
...
+        CreateWindow<WebWindow>(options).Show();
+    }
+
+    private void InjectMessageRouterConfig()
+    {
+        var server = Host.Services.GetRequiredService<IMessageRouterWebSocketServer>();
+        _logger.LogInformation($"Message Router server listening at {server.WebSocketUrl}");
+
+        WebWindow.AddPreloadScript(
+            $$"""
```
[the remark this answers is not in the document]
BalassaMarton: Hm, `LangVersion` is set to `latest` so maybe the pipeline uses old tooling?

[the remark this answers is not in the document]
BalassaMarton: OK so C# 11 is included in .NET 7 SDK, so we can use that SDK in the pipeline, or I can revert this to using the old syntax.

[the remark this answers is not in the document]
BalassaMarton: Setting `LangVersion` to `preview` solved the problem.

psmulovics:
"Chrome 110.0.0.0 (Windows 10) DatagridComponent it should create app FAILED
	Error: Unable to create the MessageRouter client, configuration is missing." from the build

BalassaMarton:
> "Chrome 110.0.0.0 (Windows 10) DatagridComponent it should create app FAILED Error: Unable to create the MessageRouter client, configuration is missing." from the build

Yup, forgot to run those tests. I'll fix the tests, `MessageRouter` should be injected and mocked instead of creating it in each component.

Tryouts/Plugins/ApplicationPlugins/datagrid/package.json
```diff
@@ -20,7 +20,7 @@
     "@angular/platform-browser": "^15.1.0",
     "@angular/platform-browser-dynamic": "^15.1.0",
     "@angular/router": "^15.1.0",
-    "material-icons": "1.13.1",
```
kruplm, line 23: We need material icons

BalassaMarton: Fixed.

Tryouts/Messaging-JS/package.json
```diff
@@ -1,25 +1,36 @@
...
+    "type": "module",
+    "files": [
+        "dist/esm/**/*"
+    ],
+    "main": "dist/esm/index.js",
+    "module": "dist/esm/index.js",
+    "types": "dist/esm/index.d.ts",
     "scripts": {
-        "build": "yarn run build:clean && yarn run build:esm && yarn run build:cjs",
+        "build": "yarn && yarn run build:clean && yarn run build:rollup",
```
kruplm, line 13: if you want to use yarn then it should be

- a) added as a devDependency
- b) invoked with npx

Unless you do that everyone in the dev team has to globally install yarn first

kruplm (changes requested):

Tryouts/Messaging-JS/package.json
BalassaMarton: Understood, removing the dependency on yarn

kruplm: ok. So delete yar.lock from Messaging-JS
and the build script should look like:  "build": "npm run build:clean && npm run build:rollup",

kruplm (approved):

PR #162, opened 2023-02-24
Reactive layout when resizing browser window

lilla28:
Responsive layout:

- As creating a responsive design, it is allowed that the font size and table will scale up or down as the browser window is resized.


kruplm (approved):
LGTM

PR #163, opened 2023-02-27
Add option to set app icon from node-launcher

ZKRobi:


Tryouts/Prototypes/Shell/Node-launcher/Lib/src/WindowConfig.ts
```diff
@@ -1,6 +1,7 @@
 export interface WindowConfig {
     title?: string;
-    url?: string;
+    icon?: string;
+    url: string;
```
BalassaMarton, line 4: You can create a `BrowserWindow` without `url` and then call `loadUrl`.

BalassaMarton (changes requested):

Tryouts/Prototypes/Shell/Node-launcher/Lib/src/WindowConfig.ts
ZKRobi: True, fixed

kruplm (approved):
LGTM

PR #173, opened 2023-03-10
Move the shell and node-launcher under src/

BalassaMarton:


BalassaMarton:
Also renamed the shell executable to ComposeUI-Shell to avoid any conflict with other apps

kruplm (approved):

PR #174, opened 2023-03-11
Add CommandLineParser tests

ZKRobi:
Added tests for CommandLineParser
Changed WebViewOptions to handle default values by itself

### DIFF
--- a/src/shell/dotnet/Shell/WebWindowOptions.cs
+++ b/src/shell/dotnet/Shell/WebWindowOptions.cs
@@ -6,7 +6,7 @@ namespace Shell
     {
         [Display(Description = "Set the height of the window. Default: 450")]
         public double? Height { get; set; }
-        
+
         [Display(Description = $"Set the title of the window. Default: {DefaultTitle}")]
         public string? Title { get; set; }
 
--- a/src/shell/dotnet/Shell.sln
+++ b/src/shell/dotnet/Shell.sln
@@ -11,6 +11,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MorganStanley.ComposeUI.Mes
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MorganStanley.ComposeUI.Messaging.Server", "..\..\messaging\dotnet\src\Server\MorganStanley.ComposeUI.Messaging.Server.csproj", "{F0DFAB55-14F5-4FF6-A12F-37B1745D4551}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tests", "tests", "{BEA71FE5-7EC8-4BEA-819B-C0DB33125D91}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Shell.Tests", "tests\Shell.Tests\Shell.Tests.csproj", "{70491F36-A27B-4E8C-AFD5-B49480EC1343}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -29,6 +33,10 @@ Global
 		{F0DFAB55-14F5-4FF6-A12F-37B1745D4551}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{F0DFAB55-14F5-4FF6-A12F-37B1745D4551}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{F0DFAB55-14F5-4FF6-A12F-37B1745D4551}.Release|Any CPU.Build.0 = Release|Any CPU
+		{70491F36-A27B-4E8C-AFD5-B49480EC1343}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{70491F36-A27B-4E8C-AFD5-B49480EC1343}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{70491F36-A27B-4E8C-AFD5-B49480EC1343}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{70491F36-A27B-4E8C-AFD5-B49480EC1343}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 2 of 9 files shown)


src/shell/dotnet/Shell/WebWindowOptions.cs
```diff
@@ -4,20 +4,40 @@ namespace Shell
...
-        public double? Height { get; set; }
-        
-        [Display(Description = $"Set the title of the window. Default: {DefaultTitle}")]
-        public string? Title { get; set; }
-
-        [Display(Description = $"Set the url for the web view. Default: {DefaultUrl}")]
-        public string? Url { get; set; }
+        public double? Height
+        {
+            get { return _height ?? DefaultHeight; }
```
BalassaMarton, line 11: I can't see the value in doing this.  Nothing will _write_ `null` to these properties, so you could just add an initializer that sets the default. You just added logic and a lot of boilerplate code to a DTO that (as discussed before) should be in sync with the `BrowserWindow` type on the JS side, which is also a dumb DTO that doesn't apply any defaults. `WebWindowOptions` is going to be part of the public API (as `BrowserWindow`), while `WebWindow` is not, that's why the defaults are declared here. This is merely to inform the consumers of this API about the defaults. And the whole pattern is misleading: the type has nullable properties, but there's hidden behavior that makes them always return a non-null value (hence you have to suppress all the warnings in the `WebWindow` constructor). I can only reiterate that this is 1. a DTO that should not contain any logic, 2. it will be part of the public API so it has to be predictable without reading the source code.

BalassaMarton (changes requested):

src/shell/dotnet/Shell/WebWindowOptions.cs
ZKRobi: I highly disagree with the statement that this should be only a dumb dto. If there is a default value defined, the options class should actually default to it, and not expect the programmer to not forget about it in other places.
The properties are nullable because otherwise they ARE initialized to the default value of the type, overriding anything you set in the constructor. This is one of the reasons I asked why do we even need to create a generic command line parser that handles default values worse than just simply configuring the normal System.CommandLine parser.

src/shell/dotnet/tests/ShellTests/ShellTests.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
ZKRobi: Done

src/shell/dotnet/Shell/WebWindow.xaml.cs
```diff
@@ -59,7 +59,7 @@ private void TrySetIconUrl(WebWindowOptions webWindowOptions)
             return;
 
         // TODO: What's the default URL if the app is running from a manifest? We should probably not allow relative urls in that case.
```
kruplm, line 61: If for any reason the URL is not set, then we're defaulting to about:blank.

src/shell/dotnet/Shell/WebWindowOptions.cs
BalassaMarton: I'm just repeating myself: this DTO should be the same as [`WindowConfig`](https://github.com/morganstanley/ComposeUI/blob/main/src/shell/js/composeui-node-launcher/src/WindowConfig.ts), it should have the same properties (mostly optional/nullable) with the same type and the same behavior (which is nothing). There's no programmer who could forget anything, and there are no other places. The only place where this type is used is the `WebWindow` constructor. Reminder: the `WebWindowOptions` class will be deserialized from JSON at some point, that was the motivation behind adding a generic command line parser and removing the old code that was building a CLI command 'by hand' (https://github.com/morganstanley/ComposeUI/commit/ba796709acd624ed3810e3e3363476046466263c#diff-076e476f12341fb17910e5cbcaaff6b3e5ea512f6020a9a2db2c4b0b1f03449a). So this object can come from command line arguments OR JSON. It would be awkward if you had a type with optional properties, you serialized it in the client, and on the other side it came out with altered properties. And again, the defaults are declared on `WebWindowOptions` because that's part of a public API, while the window itself is not. The constants are there to inform the developer of the defaults that will be applied, not to say 'even though `X` is a nullable property, it will have this value if you don't set it or set it to `null`'.
Regardless, you're right in that we should fix the command line parser, it should not set properties that have no corresponding argument. That would make it equivalent to a JSON deserializer which also doesn't change properties that are not in the JSON input. What we don't want is to constantly update the command definition when `WebWindowOptions` gets a new property. The other reason I added this generic parser is that `WebWindowOptions` is not the only thing that could get deserialized. There's another mode the shell can be started and it's when it reads a manifest. At the moment the manifest file name is hard coded, but that's obvously a TODO item before any release. Yes, we could use `System.CommandLine` and build a proper CLI IF this was a CLI app with many commands, and IF we didn't want to also deserialize these objects from other sources - but that's not the case.

ZKRobi: I agree with fixing the default handling in the serializer either by having e.g. a defaultvalue attribute, or just not overwriting a preset value if it's not provided. I couldn't find a simple way to do it, and it's not the scope of my work here, that's why I went with this pattern.
We can discuss the rest in person, not much point in repeating the same thing again and again.

src/shell/dotnet/Shell/WebWindow.xaml.cs
ZKRobi: Not my code :)

src/shell/dotnet/Shell/WebWindowOptions.cs
ZKRobi: To sum up our conversation, I will revert this implementation of the default values, and we will improve on the commandlineparser to handle the same attributes as the JSON serializer will when we implement it with support for multiple application launching. Defaults will be handled by setting them in the constructor, and not changing the value if it's not present in the input.

BalassaMarton: Defaults will NOT be set on this object. We just remove the constants and not document any defaults. The `WebWindow` constructor will apply defaults in an unspecified way (possibly with customization options later).

BalassaMarton (dismissed):

src/shell/dotnet/Shell/WebWindowOptions.cs
ZKRobi: :sigh: I guess we will need to discuss about this further.

kruplm (dismissed):
LGTM.

kruplm (approved):

PR #175, opened 2023-03-13
Move composeui-example-grid and composeui-example-chart to examples/

BalassaMarton:


BalassaMarton:
Note: there's now a `launchSettings.json` committed into the Shell project, for commonly used launch profiles.

src/shell/dotnet/Shell/Properties/launchSettings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "profiles": {
+    "Shell": {
+      "commandName": "Project"
+    },
+    "Chart and grid": {
+      "commandName": "Project",
+      "commandLineArgs": "--url http://localhost:4200"
```
kruplm, line 8: Does this result in attempting to serve both web apps (chart & grid) on the same port?

kruplm (changes requested):

src/shell/dotnet/Shell/Properties/launchSettings.json
BalassaMarton: No, this is starting the Shell assuming the grid and chart examples are running.

kruplm (approved):

PR #182, opened 2023-03-30
Process Explorer -- Added Grpc Server

lilla28:
There are lot of changes -> lot of them are : moving components to other folder, adjust/changing namespaces (as I renamed the projects too).

Main changes: 
- Removed one layer from the ProcessExplorer.Core: Now there is no ProcessMonitor, just ProcessInfoManager(abstract class: should implement for Linux and MAC(there is no option except using C++ libraries yet as I found from resources so far) still).
-  Collecting just PIDs instead of storing and changing all the time the items in it. So if there is an update for a PID, then before the sending through the server (UIHandler), we are collecting the OS specific data and then we without storing them we are sending to the UI's.
- Added SubsystemLauncher without depending on the ModuleLoader (user should set the behavior for starting, restarting or stopping a process).
- Added DI configuration possibility with tha ability to create the own server implementation.
-  Handled ObservabilityCollection.Subscribe to put the subsystem state changes into a queue (FIFO), and when a UI is connected (we are running a background task to watch the queue and the handlers), then we are taking the items out of it and updating the subsystems appropriately.

Added EndToEndTests, some Unit tests.

Changes per commits:

1.   ModulesPrototype example program fix, organization of logging, added GRpcProcessExplorer server, renamed variables, implemented IDisposable for IProcessInfoAggregator and IProcessMonitor instances:
-  Added the ProcessExplorer setup DI extensions to the Program.cs file, 
-  Added method to the IProcessMontior to set the handlers, which will handle the events from the ProcessMonitor if a process created/terminated/modified
-  Implemented IDisposable for ProcessInfoManager, modified collection to ObservableCollection which will store just the processIds, and sets the action which should be taken when the relevant processes are created/terminated/modified. Removed some unnecessary methods, added some null checks inside the implementation
-  Adjust namespaces for multiple files
-  ISubsystemLauncher API changes: Task instead of ValueTask
-  SubsystemInfo: added helper static method to convert Module object to SubsystemInfo instance.
- ProcessInfoAggregator: renamed some variables, null checks
- Some whitespace changes through the files
- Added gRPC and WebSocket Process Explorer Server implementation: initial

2.  Initial commit of tests, adjust namespaces, handle subsystem's status change, null checks for server implementation, delegating launch/stop/restart request:
- ModulesPrototype: handling subsystem's status change: putting into a queue then after everything setup we handle it, by running a background task to watch the queue and push the messages if available.
- ProtoConvertHelper: this class is helper class, which converts the objects, which should be sent to the UI to the applicable object type (Message), when using gRPC. If we won't want to extend the server implemntations by adding for example WebSocket, etc. then this could be leaved behind, and also just using the classes declared in the .proto file.
- Added ConfigureSubsystemLAuncher, which takes 2 actions and 2 functions, which creates the relevant types to start and launch subsystems, also the exact methods which would start and stop subsystems. 
- Added SubsystemLAuncher implementation to the Core package, as we should delegate now the actions, and ctors.
- Removes unneccessary files.

3. Deleting SuperRpc server POC, creating the structure of the repository's rules for Process Explorer:
- Removed SuperRpc POC solution,
- Removed LocalColelctor usage from WpfDataGrid
- Removed unneccessary files from ModulesPrototype example program, which was related to the Process Explorer
- Moved files under the relevant structure under the TryOuts/Plugins/ApplicationPlugins/process explorer/dotnet-js folders.

4. Renaming  variables, arguments, server disposing fix, end-to-end tests initial, added IProcessManager, removed unnecessary files, changed ISubsystemLauncher and communicator API's
- Renaming some methods
- Added properties to the IProcessInfoAggregator, renamed some methods, and reorganized the structure of this interface.
- Renamed some variables/arguments for better uderstanding: pid -> processId or ppid -> parentProcessId
- API changes for ISubsystemLauncher which responsible for delegating the launch/stop requests for subsystems.
- Removed prevously defined WebSocket solution as discussed.
- Added SubscriptionIsAlive update message for IUiHandler to inform the connected uis that the subscription is alive.
- Added some tests with FluentAssertions package
- Added interface for IProcessInfoManager (later IProcessMonitor)

5. Remove unnecessary methods, rename IProcessInfoManager to IProcessMonitor, null checks, added IClientConnection for tracking the clients, added WindowsProcessMonitorTests, SubsystemControllerTests, SubsystemLauncherTests, fix of EndToEndTests, added sample programs for test cases, remove unnecessary methods, renamed variables, factory method
- Redesigned IUiHandler: to decrease the CPU usage for sending messages to the uis, and not generating the messages for every connected ui, I added an AddConnection and RemoveConnection to the interface, so now it have to store the connected clients and act .
- Renamed IProcessInfoManager to IProcessMonitor and the derived classes.
- SubsystemLauncher: checking in the Restart method if the stop request was successful.

6. Fix CI jobs and tests
- CI jobs dotnet test were taking the executables and files from the Debug folder -> added configuration: Release as we are building the files into the Release folder

7. Using Assert methods instead of FluentAssertions package, npm i on ProcessExplorer frontend
8. Remove LocalCollector dependency from Abstractions and Core package, moved the interfaces and entities under Abstraction package from LocalCollector.
Fix EndToEndTests
- Code Coverage report was not working with the Ecxeption: The process is being used by an anouther process. (LocalCollector.dll) -> Removed the dependency, reorganized the classes into the Abstractions project, later will revisit the LocalCollector project.

10. Fix SetProcessIds test
- Added Thread.Sleep method to wait for sure until the test application will start.

11. ClientSubscription potential fix
- Disposing the call properly


.github/workflows/continuous-integration-prototypes.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       run: Get-ChildItem -Recurse -Include *.sln | ForEach-Object {dotnet build $_ --configuration Release --no-restore; if ($LASTEXITCODE -ne 0 ) {throw "Build for $_ FAILED"; }} 
 
     - name: Test .Net
-      run: Get-ChildItem -Recurse -Include *.sln | ForEach-Object {dotnet test $_ --no-restore --verbosity normal --collect:"XPlat Code Coverage"}  
+      run: Get-ChildItem -Recurse -Include *.sln | ForEach-Object {dotnet test $_ --configuration Release --no-restore --verbosity normal --collect:"XPlat Code Coverage"}
```
kruplm, line 35: Nice catch! :)

kruplm: You might need to PR this one liner separately though because github actions on the main repo are always executed based on what is merged into the main repo's master branch.

kruplm: Sorry, never mind, I checked what's being executed for the PR.
It seemed to pick up the changes you've pushed.
There are other circumstances where you need to fix the workflow separately but here it seems to take effect already.

ZKRobi (changes requested):

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/ProcessExplorer.Core/ProcessInfoAggregator.cs
```diff
@@ -0,0 +1,386 @@
...
+
+        if (SubsystemController != null)
+            SubsystemController.SetUiDelegate(UpdateInfoOnUI);
+
+        SetUiCommunicatorsToWatchProcessChanges();
+    }
+
+    private void DisposeCore()
+    {
+        _processInfoMonitor.Dispose();
```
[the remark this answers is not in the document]
lilla28: You are totally right, missed that! Actually as I see there is no need to implement IDisposable for IProcessInfoAggregator class at all at this point. :)

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/ProcessExplorer.Core/Processes/WindowsProcessMonitor.cs
```diff
@@ -0,0 +1,350 @@
...
+        _memoryPerformanceCounters.TryRemove(processId, out var memoryPerf);
+        memoryPerf?.Close();
+        memoryPerf?.Dispose();
+
+        _managementPidObjectSearchers.TryRemove(processId, out var objectSearcher);
+        objectSearcher?.Dispose();
+
+        lock (_lock)
+        {
+            var pidExists = ContainsId(processId);
```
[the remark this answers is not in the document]
lilla28: It checks if the the processId exists in the collection, which stores the processIds what are being monitored by the ProcessMonitor implementation.

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/LocalCollector/ProcessInfoCollector.cs
```diff
@@ -11,97 +11,80 @@
...
+using ProcessExplorer.Abstractions.Entities;
+using ProcessExplorer.Abstractions.Entities.Connections;
+using ProcessExplorer.Abstractions.Infrastructure;
 
 namespace LocalCollector;
 
 public class ProcessInfoCollector : IProcessInfoCollector
 {
-    public ProcessInfoCollectorData Data { get; } = new();
+    public ProcessInfoCollectorData ProcessInformation { get; } = new();
```
ZKRobi, line 29: As you are using this inside lock blocks, it should be private, otherwise external code can cause concurrency issues

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/LocalCollector/ProcessInfoCollector.cs
```diff
@@ -162,25 +145,23 @@ public async Task AddConnectionMonitor(ConnectionMonitorInfo connections)
...
     {
-        await AddConnectionMonitor(connections.Data);
+        await AddConnectionMonitor(connections.Connections);
     }
 
-
     public async Task AddEnvironmentVariables(EnvironmentMonitorInfo environmentVariables)
     {
-        if (Data.EnvironmentVariables.IsEmpty)
+        if (ProcessInformation.EnvironmentVariables.IsEmpty)
```
ZKRobi, line 160: The check and the operation should be both inside the lock block

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/LocalCollector/ProcessInfoCollector.cs
```diff
@@ -210,7 +191,7 @@ public async Task AddEnvironmentVariables(EnvironmentMonitorInfo environmentVari
         SynchronizedCollection<T> source,
         SynchronizedCollection<T> target,
         Func<T, Func<T, bool>> predicate,
-        Func<IEnumerable<KeyValuePair<AssemblyInformation, IEnumerable<T>>>, ValueTask> handler)
+        Func<IEnumerable<KeyValuePair<RuntimeInformation, IEnumerable<T>>>, ValueTask> handler)
     {
         if (!target.Any())
```
ZKRobi, line 196: Same here, SynchronizedCollections are thread-safe on a single operation, but between 2 operations, they can change

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/ProcessExplorer.Abstractions/Processes/ProcessInfoMonitor.cs
```diff
@@ -0,0 +1,217 @@
...
+    /// </summary>
+    /// <returns></returns>
+    public abstract float GetCpuUsage(int processId, string processName);
+
+    /// <summary>
+    /// Continuously watching created processes.
+    /// </summary>
+    public virtual void WatchProcesses(int mainProcessId)
+    {
+        ClearProcessIds();
```
ZKRobi, line 138: I'd add a lock around this method as well, to avoid a scenario like:
thread1 ClearProcessIds;
thread1 AddProcess
thread2 ClearProcessIds
thread1AddChildProcesses;

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/LocalCollector/ProcessInfoCollector.cs
```diff
@@ -157,103 +141,81 @@ public async Task AddConnectionMonitor(IConnectionMonitor connections)
...
-                }
+                _processInformation.EnvironmentVariables.AddOrUpdate(env.Key, env.Value, (_, _) => env.Value);
             }
 
-        }
-
-        if (_communicator != null)
-        {
-            var info = new List<KeyValuePair<RuntimeInformation, IEnumerable<KeyValuePair<string, string>>>>()
+            info = new List<KeyValuePair<RuntimeInformation, IEnumerable<KeyValuePair<string, string>>>>()
```
ZKRobi, line 153: Just a hint, no need to change it, a List of KeyValuePairs is essentially a Dictionary :)

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/LocalCollector/ProcessInfoCollector.cs
```diff
@@ -157,103 +141,81 @@ public async Task AddConnectionMonitor(IConnectionMonitor connections)
...
             {
                 new(_runtimeId, environmentVariables.EnvironmentVariables)
             };
+        }
 
+        if (info.Any())
```
ZKRobi, line 159: Again, I don't think it's necessary to change, just a small observation that info is never empty here - it's change-proof to still check it though.

ZKRobi (dismissed):

Tryouts/Plugins/ApplicationPlugins/MorganStanley.ComposeUI.ProcessExplorer/dotnet/src/LocalCollector/ProcessInfoCollector.cs
lilla28: oopsie, you are right, I will update the PR. just FYI I will a little bit refactor the whole LocalCollector anyway in the next 2 iteration.

ZKRobi (approved):

PR #189, opened 2023-04-14
Create Archive [WIP]

kruplm:


.github/workflows/release.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  build:
+    name: Upload Release Asset
+    runs-on: windows-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Build project # This would actually build your project, using zip for an example artifact
+        run: |
+          Compress-Archive -WhatIf -Path D:\a\ComposeUI\ComposeUI\src\shell\dotnet\Shell\bin\Debug\net6.0-windows\* composeui_win32.zip
```
BalassaMarton, line 18: Don't assume the workspace directory, use `${{ github.workspace }}/src/...` instead. In general, I would avoid using `\` as path separator since it's only supported on Windows, whereas `/` works on any platform.

psmulovics (approved):

BalassaMarton (approved):

BalassaMarton (approved):

PR #206, opened 2023-05-02
Update deploy-to-npm.yml

kruplm:


.github/workflows/deploy-to-npm.yml
```diff
@@ -19,16 +19,9 @@ jobs:
...
+      - run: npm run build
+      - run: npm publish
       #- run: npm config set //registry.npmjs.org/:_authToken=$NPM_TOKEN
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-
-      - name: Publish composeui-node-launcher
-        run: npm publish
         working-directory: ${{ github.workspace }}/src/shell/js/composeui-node-launcher/
```
bingenito, line 27: Is this an ok working dir for all the above commands?  Is that why npm publish would be ok?

bingenito, line 27: Is this an ok working dir for all the above commands? Is that why npm publish would be ok?

.github/workflows/deploy-to-npm.yml
```diff
@@ -19,16 +19,9 @@ jobs:
           node-version: ${{ matrix.node-version }}
           registry-url: 'https://registry.npmjs.org'
       - run: npm ci
-      - run: npx lerna run build
-      #- run: npm publish
+      - run: npm run build
+      - run: npm publish
```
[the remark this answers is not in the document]
kruplm: I attempt to publish this directly:
https://github.com/morganstanley/ComposeUI/blob/main/src/shell/js/composeui-node-launcher/package.json 

To see if it works.

.github/workflows/deploy-to-npm.yml
kruplm: yes it should be ok for all the other commands.
yes, it ignores the fact that this is a lerna monorepo, but it should work.

PR #219, opened 2023-05-10
Fixing typo: executeScrtiptFile -> executeScriptFile

ztanczos:
Fixing typo


kruplm (changes requested):
Fix the typo in the spec too. Thanks.

PR #221, opened 2023-05-22
Adjust namespaces, moving under prototypes folder, removed Tryouts

lilla28:
Adjust namespaces, moving under prototypes folder, removed Tryouts, edited lerna config, moved defaultApp under prototypes, unnecessary recreation of Module in ModulesPrototypes fix - state modification, after renaming the package fixing tests


kruplm (approved):
LGTM

package.json
```diff
@@ -6,7 +6,7 @@
     "examples/js-*",
     "src/messaging/js/*",
     "src/shell/js/*",
-    "Tryouts/Plugins/ApplicationPlugins/*"
+    "prototypes/*"
```
kruplm, line 9: I just noticed on the console, that the project doesn't seem to be picked up under the new location.
Could you change this to `prototypes/process-explorer/js` ?
If it's still not picked up, we can merge ad I'm looking at it locally.

kruplm (changes requested):

kruplm (approved):

kruplm (approved):

PR #227, opened 2023-05-30
Added skeleton project for fdc desktop agent TS implementation

ZKRobi:
Also some minor consistency changes

### DIFF
--- a/src/shell/js/composeui-fdc3/tsconfig.json
+++ b/src/shell/js/composeui-fdc3/tsconfig.json
@@ -1,0 +1,14 @@
+﻿{
+"compilerOptions": {
+    "moduleResolution": "node",
+	"types": [ "node" ],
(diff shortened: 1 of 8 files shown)


src/shell/js/composeui-fdc3/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+﻿{
+"compilerOptions": {
+    "moduleResolution": "node",
```
kruplm, line 3: including "types": [ "node", "jest" ], here should resolve the current issue on the CI regarding types:
see: https://github.com/morganstanley/ComposeUI/blob/main/src/shell/js/composeui-node-launcher/tsconfig.json

kruplm (changes requested):

psmulovics (approved):

PR #236, opened 2023-06-13
Initial Fdc3 implemetation using MessageRouter, added unit tests

lilla28:


src/shell/js/composeui-fdc3/tsconfig.json
```diff
@@ -1,14 +1,14 @@
...
-    "esModuleInterop": true,
-    "target": "es2022",
-    "declaration": true,
-    "outDir": "output",
-    "lib": [ "es2022" ],
-    "strict": true
+    "types": [ "node", "jest" ],
+      "module": "commonjs",
+      "esModuleInterop": true,
+      "target": "es6",
```
ztanczos, line 7: Why the downgrade?

src/shell/js/composeui-fdc3/src/ComposeUIListener.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export class ComposeUIListener implements Listener {
+    private messageRouterClient: MessageRouter;
+    private unsubscribable?: Unsubscribable;
+    private handler?: ContextHandler;
+    private channelId?: string;
+    private contextType?: string;
+    private isSubscribed: boolean = true;
+
+    constructor (messageRouterClient: MessageRouter, handler?: ContextHandler, channelId?: string, contextType?: string) {
```
ztanczos, line 27: If a ContextHandler is required for every operation why is it marked as optional here?

src/shell/js/composeui-fdc3/src/ComposeUIListener.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.handler = handler;
+
+        if(channelId != null) { //topic: "composeui/fdc3/v2.0/userchannels/{name}/"
+            this.channelId = channelId;
+        }
+        
+        this.contextType = contextType;
+    }
+
+    public async subscribe(channel: string): Promise<void> { //"composeui/fdc3/v2.0/userchannels/{name}/broadcast"
```
ztanczos, line 38: Can someone call `subscribe` more than once? E.g.: for different channels. In that case we'd need more than one `unsubscribable`

lilla28: In the current implementation no, as per I assume we would like to export just the DesktopAgent implementation, which would call the listener's subscribe method with the suffix "broadcast" in the addContextListener method. I am not really sure if that's the right approach for the fdc3 though.

src/shell/js/composeui-fdc3/src/Fdc3ComposeUI.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+    it('broadcast will set the lastContext to test instrument', async() => {
+        const testChannel = new ComposeUIChannel(dummyTopic, ComposeUIChannelType.User, messageRouterClient);
+        await testChannel.broadcast(testInstrument);
+        const resultContext = await testChannel.getCurrentContext();
+        const expectedObject = {
+            Id: dummyTopic + "broadcast",
+            Context: testInstrument
+        };
+        expect(messageRouterClient.mock.publish).toHaveBeenCalledTimes(1);
+        expect(messageRouterClient.PublishedMessages.size).toBe(1);
```
BalassaMarton, line 55: Just check that `publish` was called correctly using jest's built-in features:
```ts
expect(messageRouterClient.mock.publish).toHaveBeenCalledWith(dummyTopic + "broadcast", JSON.stringify(expectedObject));
```

src/shell/js/composeui-fdc3/src/Fdc3ChannelMessage.ts
```diff
@@ -13,8 +13,8 @@
 
 import { Context } from "@finos/fdc3";
 
-export class Fdc3ChannelMessageRouterMessage {
-    public Id!: string; //this is the topic actually
+export class Fdc3ChannelMessage {
+    public Id!: string;
```
BalassaMarton, line 17: JS member names should be camelCase

src/shell/js/composeui-fdc3/src/ComposeUIChannel.ts
```diff
@@ -0,0 +1,69 @@
...
+import { MessageRouter } from "@morgan-stanley/composeui-messaging-client";
+import { ChannelType } from "./ChannelType";
+import { ComposeUIListener } from "./ComposeUIListener";
+import { Fdc3ChannelMessage } from "./Fdc3ChannelMessage";
+import { ComposeUITopic } from "./ComposeUITopic";
+
+export class ComposeUIChannel implements Channel{
+    id!: string;
+    type!: "user" | "app" | "private";
+    displayMetadata?: DisplayMetadata | undefined;
```
BalassaMarton, line 24: @kruplm what's your opinion on this? I think an optional property or parameter of type `T?` implicitly has the type `T | undefined`, so you don't need `| undefined`.

src/shell/js/composeui-fdc3/src/ComposeUIChannel.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    public broadcast(context: Context): Promise<void> {
+        //Setting the last published context message.
+        this.lastContexts.set(context.type, context);
+        this.lastContext = context;
+        const fdc3Message = new Fdc3ChannelMessage(this.id, context);
+        return this.messageRouterClient.publish(ComposeUITopic.broadcast(this.id), JSON.stringify(fdc3Message));
+    }
+
+    public getCurrentContext(contextType?: string | undefined): Promise<Context | null> {
```
BalassaMarton, line 44: Similarly, `contextType?: string` already means its type is `string or undefined` and you can check that it has an actual value by just coercing it into a truth value:
```ts
if (contextType) {
    // contextType is neither null, undefined or empty string
}
```

src/shell/js/composeui-fdc3/src/ComposeUIChannel.ts
```diff
@@ -0,0 +1,69 @@
...
+        if (contextType != null && contextType != undefined) {
+            return new Promise<Context>((resolve, reject) => {
+                console.log("Resolving the current context: ", contextType);
```
BalassaMarton, line 47: I'd add some // TODO comments to remove the logging once the PR is ready (I assume these are for debugging)

src/shell/js/composeui-fdc3/src/ComposeUIChannel.ts
```diff
@@ -0,0 +1,69 @@
...
+                resolve(this.lastContext!);
+            });
+        }
+    }
+
+    public addContextListener(contextType: string | null, handler: ContextHandler): Promise<Listener>;
+    public addContextListener(handler: ContextHandler): Promise<Listener>;
+    public async addContextListener(contextType: any, handler?: any): Promise<Listener> {
+        if(typeof contextType != 'string'){
+            throw new Error("addContextListener without contextType is depracted, please use the newer version.");
```
BalassaMarton, line 62: Typo: 'depracted'

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    public getUserChannels(): Promise<Array<Channel>> {
+        return Promise.resolve(this.userChannels);
+    }
+
+    public joinUserChannel(channelId: string): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            let channel = this.userChannels.find(innerChannel => innerChannel.id == channelId);
+            if (!channel) {
+                reject(new Error("Channel couldn't be found in user channels"));
```
BalassaMarton, line 124: Reject with the error message defined in the standard, see https://github.com/finos/FDC3/blob/master/src/api/DesktopAgent.ts#LL386C24 and https://github.com/finos/FDC3/blob/master/src/api/Errors.ts#L48
Add a TODO to do a sweep for non-standard errors.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,254 @@
...
+                resolve();
+            }
+        });
+    }
+
+    public getOrCreateChannel(channelId: string): Promise<Channel> {
+        return new Promise<Channel>((resolve, reject) => {
+            let channel = this.userChannels.find(innerChannel => innerChannel.id == channelId);
+            if (!channel) {
+                channel = new ComposeUIChannel(channelId, "app", this.messageRouterClient); //TODO later
```
BalassaMarton, line 136: First it looks for the channel id in `userChannels`, but the new channel is added to `appChannels`, is that intentional?

BalassaMarton: OK it has to be an app channel, so the find must be in `appChannels`

src/shell/js/composeui-fdc3/src/ComposeUIListener.ts
```diff
@@ -0,0 +1,68 @@
...
+    private handler: ContextHandler;
+    private channelId?: string;
+    private contextType?: string;
+    private isSubscribed: boolean = false;
+
+    constructor(messageRouterClient: MessageRouter, handler: ContextHandler, channelId?: string, contextType?: string) {
+        this.messageRouterClient = messageRouterClient;
+        this.handler = handler;
+
+        if (channelId != null) {
```
BalassaMarton, line 32:
```ts
if (channelId)
```
If I call this constructor with just the two mandatory parameters, the `channelId` and `contextType` will both be `undefined` (not `null`)

src/shell/js/composeui-fdc3/src/ComposeUITopic.ts
```diff
@@ -0,0 +1,25 @@
...
+ *  and limitations under the License.
+ *  
+ */
+
+export class ComposeUITopic {
+    private static readonly topicRoot = "composeui/fdc3/v2.0/";
+    private static readonly userChannelTopicRoot: string = this.topicRoot + "userchannels/";
+
+    static broadcast(topicId: string) {
+        return this.userChannelTopicRoot + topicId + "/broadcast";
```
BalassaMarton, line 19: Isn't broadcast channel-specific? Shouldn't `broadcast` and `subscribe` return the same topic for a given `channelId` and `contextType`?

src/shell/js/composeui-fdc3/src/ComposeUIChannel.ts
lilla28: I can remove the undefined, the IDE did add that automatically :D

BalassaMarton: That's why I added @kruplm , I'm not 100% sure my assumption is correct

src/shell/js/composeui-fdc3/src/ComposeUITopic.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  to in writing, software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ *  or implied. See the License for the specific language governing permissions
+ *  and limitations under the License.
+ *  
+ */
+
+export class ComposeUITopic {
+    private static readonly topicRoot = "composeui/fdc3/v2.0/";
+    private static readonly userChannelTopicRoot: string = this.topicRoot + "userchannels/";
```
kruplm, line 16: The concatenation of strings is correct, but there's a way that makes them more readable in the long run, since a topic is is basically substring1/substring2/...

I'd remove the closing / from the end of topicRoot and construct the formatted strings the following way with Template literal strings:

`` private static readonly userChannelTopicRoot: string = `${topicRoot}/userchannels/` ``

same for line 19.

Optionally you can also create const variables for the "userchannels" and "broadcast" strings, the readability could improve even more, and all the constants would be in one place, but only if we'll reuse those:

`${topicRoot}/${USER_CHANNELS}/`

I'd also put "v2.0" in a variable too for future improvements.

kruplm (changes requested):

src/shell/js/composeui-fdc3/src/ComposeUIChannelType.ts
```diff
@@ -0,0 +1,18 @@
...
+ *  See the NOTICE file distributed with this work for additional information
+ *  regarding copyright ownership. Unless required by applicable law or agreed
+ *  to in writing, software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ *  or implied. See the License for the specific language governing permissions
+ *  and limitations under the License.
+ *  
+ */
+
+export enum ComposeUIChannelType {
```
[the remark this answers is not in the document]
bingenito: Interestingly it doesn't seem to be separately defined or exported as a type.  The type property on Channel itself is just defined as the string enum.

https://github.com/finos/FDC3/blob/master/src/api/Channel.ts#L35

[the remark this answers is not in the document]
BalassaMarton: That's actually why I asked if we can contribute to Finos, I want to add that type :)

src/shell/js/composeui-fdc3/src/infrastructure/messages/Fdc3ErrorResponse.ts
```diff
@@ -0,0 +1,22 @@
...
+ *  See the NOTICE file distributed with this work for additional information
+ *  regarding copyright ownership. Unless required by applicable law or agreed
+ *  to in writing, software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ *  or implied. See the License for the specific language governing permissions
+ *  and limitations under the License.
+ *  
+ */
+
+export class Fdc3ErrorResponse {
```
BalassaMarton, line 14: Isn't this suppoed to be `Fdc3FindChannelResponse`?

src/shell/js/composeui-fdc3/src/infrastructure/messages/Fdc3GetCurrentContextRequest.ts
```diff
@@ -0,0 +1,20 @@
...
+ *  regarding copyright ownership. Unless required by applicable law or agreed
+ *  to in writing, software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ *  or implied. See the License for the specific language governing permissions
+ *  and limitations under the License.
+ *  
+ */
+
+export class Fdc3GetCurrentContextRequest {
+    public readonly contextType?: string | null;
```
BalassaMarton, line 15: Does it make sense to have both `undefined` and `null` as context type? It could be just `contextType?: string`.

src/shell/js/composeui-fdc3/src/infrastructure/messages/Fdc3ContextMessage.ts
```diff
@@ -0,0 +1,24 @@
...
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ *  or implied. See the License for the specific language governing permissions
+ *  and limitations under the License.
+ *  
+ */
+
+import { Context } from "@finos/fdc3";
+
+export class Fdc3ContextMessage {
+    public readonly Id!: string;
```
BalassaMarton, line 17: JS property names should be camelCase

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    private async invokeChannelCreationMessage(topic: string, channelId: string, channelType: ChannelType): Promise<void> {
+        const message = JSON.stringify(new Fdc3FindChannelRequest(channelId, channelType));
+        await this.messageRouterClient.invoke(topic, message)
+                    .then((response) => {
+                        if(response) {
+                            const message = JSON.parse(response) as TopicMessage;
+                            if(message.payload) {
+                                const fdc3Message = JSON.parse(message.payload) as Fdc3ErrorResponse;
```
BalassaMarton, line 301: That won't work, you won't get an object that's actually an `Fc3ErrorResponse`, you can't deserialize a class like that. The `as` operator in TypeScript just tells the compiler to treat the value *as if* it was of the specified type. The response should be declared as an interface instead.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+                break;
+            case "private":
+                this.privateChannels.push(channel);
+                break;
+        }
+    }
+
+    private async invokeChannelCreationMessage(topic: string, channelId: string, channelType: ChannelType): Promise<void> {
+        const message = JSON.stringify(new Fdc3FindChannelRequest(channelId, channelType));
+        await this.messageRouterClient.invoke(topic, message)
```
BalassaMarton, line 296: Don't mix async-await with promise chaining.
```ts
const response = await this.messageRouterClient.invoke(topic, message);
const message = JSON.parse(response) as TopicMessage;
// ...
```

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+                    .then((response) => {
+                        if(response) {
+                            const message = JSON.parse(response) as TopicMessage;
+                            if(message.payload) {
+                                const fdc3Message = JSON.parse(message.payload) as Fdc3ErrorResponse;
+                                if(fdc3Message.Error) {
+                                    throw new Error(fdc3Message.Error); //Type of the message should be created.
+                                } else if (fdc3Message.Found){
```
BalassaMarton, line 304: `else` block is unnecessary when the `if` condition always throws or returns.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+
+            channel = this.findChannel(channelId, "private");
+            if (channel) {
+                this.currentChannel = channel;
+                resolve();
+                return;
+            }
+
+            if (!channel) {
+                reject(new Error(`No channel is found with id: ${channelId}`));
```
BalassaMarton, line 254: Reject with `ChannelError.NoChannelFound` instead to follow the standard.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    //TODO
+    public getAppMetadata(app: AppIdentifier): Promise<AppMetadata> {
+        throw new Error("Not implemented");
+    }
+
+    //TODO
+    public getSystemChannels(): Promise<Channel[]> {
+        throw new Error("Not implemented");
```
BalassaMarton, line 221: Return an empty array? That's a safer default implementation than throwing.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+
+    public getCurrentChannel(): Promise<Channel | null> {
+        return Promise.resolve(this.currentChannel!);
+    }
+
+    //TODO: add messageRouter message that we are leaving the current channel to notify the backend.
+    public leaveCurrentChannel(): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            this.currentChannel = undefined;
+            this.currentChannelListeners.forEach(async listener => {
```
BalassaMarton, line 188: Why is the callback async?

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+        throw new Error("Not implemented.");
+    }
+
+    //TODO
+    public createPrivateChannel(): Promise<PrivateChannel> {
+        throw new Error("Not implemented");
+    }
+
+    public getCurrentChannel(): Promise<Channel | null> {
+        return Promise.resolve(this.currentChannel!);
```
BalassaMarton, line 181: I think I made you change the type of `currentChannel`, but here I can see I was wrong: it should be `currentChannel: Channel | null`. 
Alternatively, you can resolve with `this.currentChannel ?? null` to make sure that `undefined` is never returned.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+                this.currentChannel = channel;
+                resolve();
+            }
+        });
+    }
+
+    //TODO: should return AccessDenied error when a channel object is denied
+    //TODO: should return a CreationFailed error when a channel cannot be created or retrieved (channelId failure)
+    public getOrCreateChannel(channelId: string): Promise<Channel> {
+        // return new Promise<Channel>(async(resolve, reject) => {
```
BalassaMarton, line 159: Is this not supported by the backend? Why is it commented out?

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,312 @@
...
+            if (!this.currentChannel) {
+                reject(new Error("The current channel have not been set."));
+                return;
+            }
+            if (typeof contextType != 'string' || !contextType) {
+                reject(new Error("The contextType was type of ContextHandler, which would use a deprecated function, please use string or null for contextType!"));
+                return;
+            }
+
+            const stringContextType = contextType as string ?? null;
```
BalassaMarton, line 111: Due to the previous condition, `contextType` is already narrowed to `string` here, no need for a new variable.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
lilla28: We agreed on that if that is not supported by the backend then I should comment out / remove the code. I decided to comment out.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
lilla28: Yeah it is a deprecated function which is an alias to the getUserChannels. I missed that, will update this:
https://github.com/finos/FDC3/blob/master/src/api/DesktopAgent.ts#L535-L537

src/shell/js/composeui-fdc3/src/infrastructure/messages/Fdc3FindChannelResponse.ts
```diff
@@ -0,0 +1,17 @@
...
+ *  regarding copyright ownership. Unless required by applicable law or agreed
+ *  to in writing, software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ *  or implied. See the License for the specific language governing permissions
+ *  and limitations under the License.
+ *  
+ */
+
+export interface Fdc3FindChannelResponse {
+    error: string | undefined;
```
BalassaMarton, line 15: Short form: `error?: string`

src/shell/js/composeui-fdc3/src/infrastructure/ComposeUIChannel.ts
```diff
@@ -0,0 +1,93 @@
...
+        const topic = ComposeUITopic.broadcast();
+        await this.messageRouterClient.publish(topic, JSON.stringify(message));
+    }
+
+    //TODO add error
+    public getCurrentContext(contextType?: string | undefined): Promise<Context | null> {
+        return new Promise<Context | null>(async (resolve, reject) => {
+            const message = JSON.stringify(new Fdc3GetCurrentContextRequest(contextType));
+            await this.messageRouterClient.invoke(ComposeUITopic.getCurrentContext(this.id, this.type), message)
+                .then((response) => {
```
BalassaMarton, line 52: Don't mix `async-await` with `then`

src/shell/js/composeui-fdc3/src/ComposeUITopic.ts
kruplm: Sorry, I have to nitpick, somehow the current solution looks like "concatenation with extra steps" instead of leveraging what template strings are for.

src/shell/js/composeui-fdc3/src/infrastructure/ComposeUITopic.ts
```diff
@@ -0,0 +1,58 @@
...
+ *  
+ */
+
+import { ChannelType } from "./ChannelType";
+
+//TODO: Add more methods which returns the topics for ComposeUI's right channel
+export class ComposeUITopic {
+    private static readonly fdc3Version = "v2.0";
+    private static readonly topicRoot = `$composeui/fdc3/${this.fdc3Version}`;
+    private static readonly userChannels = '/userChannels/';
```
kruplm, line 20: These strings should be without the separators so in this example `private static readonly userChannels = 'userChannels';`
And so on for the rest of the strings.

src/shell/js/composeui-fdc3/src/infrastructure/ComposeUITopic.ts
```diff
@@ -0,0 +1,58 @@
...
+    private static readonly appChannels = '/appChannels/';
+    private static readonly privateChannels = '/privateChannels/';
+    private static readonly broadcastSuffix = "/broadcast";
+    private static readonly getCurrentContextSuffix = "/getCurrentContext";
+    private static readonly joinUserChannelSuffix = "/joinUserChannel";
+    private static readonly getOrCreateChannelSuffix = "/getOrCreateChannel";
+
+    public static broadcast() {
+        return `${this.topicRoot}${this.broadcastSuffix}`;
```
kruplm, line 29: And with the updated strings without the separators you can construct this string by leveraging template strings like the following:
`` return `${this.topicRoot}/${this.broadcastSuffix}`; ``

And so on for the rest of the functions.

The benefit is, that you don't have to keep in mind where the separators should be for every subsrting.

So instead of doing substring1substring2, where you have to be aware of the placement of the separators, you can just simply do t as:
substring1/substring2

kruplm (changes requested):

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,308 @@
...
+import { ComposeUIListener } from './infrastructure/ComposeUIListener';
+import { Fdc3FindChannelRequest } from './infrastructure/messages/Fdc3FindChannelRequest';
+import { Fdc3FindChannelResponse } from './infrastructure/messages/Fdc3FindChannelResponse';
+import { ComposeUITopic } from './infrastructure/ComposeUITopic';
+
+export class ComposeUIDesktopAgent implements DesktopAgent {
+    private appChannels: ComposeUIChannel[] = [];
+    private userChannels: ComposeUIChannel[] = [];
+    private privateChannels: ComposeUIChannel[] = [];
+    private currentChannel?: ComposeUIChannel;
```
BalassaMarton, line 41: Simplify as `currentChannel?: ComposeUIChannel` and check any warnings. If something is just optional, you don't need `null` in the type spec.

src/shell/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts
```diff
@@ -0,0 +1,308 @@
...
+                return;
+            }
+            if (typeof contextType != 'string' || !contextType) {
+                reject(new Error("The contextType was type of ContextHandler, which would use a deprecated function, please use string or null for contextType!"));
+                return;
+            }
+
+            const listener = <ComposeUIListener>await this.currentChannel!.addContextListener(contextType, handler!);
+            await this.currentChannel!.getCurrentContext(contextType)
+                .then(async (resultContext) => {
```
BalassaMarton, line 112: Don't mix `async-await` with `then`.

```ts
const resultContex = await this.currentChannel!.getCurrentContext(contextType);
listener.LatestContext = this.currentChannel!.retrieveCurrentContext(contextType);
// ...

BalassaMarton (approved):

kruplm (approved):
LGTM

PR #239, opened 2023-06-20
Messaging: Added IConnectionFactory for correctly implementing disposable pattern

BalassaMarton:
Added IConnectionFactory for correctly implementing disposable pattern

### DIFF
--- a/src/messaging/dotnet/src/Client/Client/MessageRouterClient.cs
+++ b/src/messaging/dotnet/src/Client/Client/MessageRouterClient.cs
@@ -25,11 +25,11 @@ namespace MorganStanley.ComposeUI.Messaging.Client;
 internal sealed class MessageRouterClient : IMessageRouter
 {
     public MessageRouterClient(
-        IConnection connection,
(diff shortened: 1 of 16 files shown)


src/messaging/dotnet/src/Client/Client/MessageRouterClient.cs
```diff
@@ -185,7 +185,7 @@ private Topic GetTopic(string topicName)
     }
 
     private string? _clientId;
-    private readonly IConnection _connection;
+    private IConnection _connection;
```
ztanczos, line 188: Nitpick: why did you remove `readonly`?

ztanczos (dismissed):

src/messaging/dotnet/src/Client/Client/MessageRouterClient.cs
BalassaMarton: That was totally unintentional, thanks. I'll fix it in a sec.

ztanczos (approved):

PR #243, opened 2023-06-23
Preload Fdc3 2.0 ComposeUIDesktopAgent for Embedded Browser

kruplm:
Changes in this PR
* an IIFE bundle is being generated from our FDC3 DesktopAgent implementation, and preloaded into the Embedded Browser

Work in Progress 
* I intend to discuss the mechanism around copying the bundle or have an agreement on a central folder where the different components of the shell can access these kind of artifacts.
* The current PowerShell script is just simply copying the already generated bundle I'm going to change it according to what we decide.
* InjectFDC3(): It's part of the Shell, but it should be integrated into the FDC3 backend in a separate integration PR.
* The exports for the ComposeUIDesktopAgent may need to change as accessing the methods like fdc3.ComposeUIDesktopAgent.broadcast() may contradicts the standard


src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -170,5 +172,16 @@ private void InjectMessageRouterConfig()
                     
                 """);
     }
+   
+    private async void InjectFDC3()
```
BalassaMarton, line 176: `async` keyword is unnecessary, this is not an async method.

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -35,6 +35,9 @@
 		<None Update="Manifest\exampleManifest.json">
 			<CopyToOutputDirectory>Always</CopyToOutputDirectory>
 		</None>
+		<None Update="Preload\fdc3-iife-bundle.js">
```
BalassaMarton, line 38: `<Content Include="..."`
See https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2022

ztanczos: Shouldn't this be an embedded resource instead?

BalassaMarton: You're right, we already discussed and agreed to embed it.

kruplm: It's configured to be an Embedded Resource now.

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -170,5 +172,16 @@ private void InjectMessageRouterConfig()
                     
                 """);
     }
+   
+    private void InjectFDC3()
```
BalassaMarton, line 176: I don't think FDC3 should be injected by default. It should be an opt-in feature that you can enable in `WebWindowOptions`, preferably in a `Features` collection so that it can be factored out into a plugin later (without breaking existing apps).

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -170,5 +172,16 @@ private void InjectMessageRouterConfig()
...
+    {
+        string bundle = @"fdc3-iife-bundle.js";
+        string processPath = Environment.ProcessPath;
+        string folder = Path.GetDirectoryName(processPath);
+        string path = Path.Combine(folder, @"Preload\", bundle);
```
BalassaMarton, line 181: This no longer works as the bundle is now an embedded resource. The only reason you're not getting errors is because the file is there as leftover from previous builds. Use https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getmanifestresourcestream?view=net-6.0#system-reflection-assembly-getmanifestresourcestream(system-string)

kruplm: Yes, I'm currently working on it. I pushed what I had yesterday in order to rebase my branch.

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -175,4 +179,25 @@ private void InjectMessageRouterConfig()
...
+        string iife = ReadResource(bundle);
+
+        WebWindow.AddPreloadScript(iife);
+    }
+
+    public string ReadResource(string name)
+    {
+        var assembly = Assembly.GetExecutingAssembly();    
+        string resourcePath = assembly.GetManifestResourceNames()
+            .Single(str => str.EndsWith(name));
```
ZKRobi, line 195: This doesn't seem very safe, it at the minimum adds an implicit requirement to have our resources not be postfixes of each other.

src/shell/dotnet/Shell/App.xaml.cs
kruplm: Bundle is being accessed from the Embedded Resource now.

src/shell/dotnet/Shell/App.xaml.cs
kruplm: I agree, but let's do that on a separate PR.

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -175,4 +178,24 @@ private void InjectMessageRouterConfig()
...
+        string bundle = @"fdc3-iife-bundle.js";
+        string iife = ReadResource(bundle);
+
+        WebWindow.AddPreloadScript(iife);
+    }
+
+    public string ReadResource(string name)
+    {
+        var assembly = Assembly.GetExecutingAssembly();
+        string resourcePath = $$"""Shell.Preload.{{name}}""";
```
BalassaMarton, line 193: The behavior of this method doesn't match the name. `Shell.Preload` is specific to that particular resource (the FDC3 bundle). Instead, pass the full resource path in `name`.

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -175,4 +178,24 @@ private void InjectMessageRouterConfig()
...
+        
+        using (Stream stream = assembly.GetManifestResourceStream(resourcePath))
```
BalassaMarton, line 195: Use implicitly typed variables and using declarations when possible, eg.

```cs
using var stream = assembly.GetManifestResourceStream(resourcePath);
using var reader = new StreamReader(stream);

return reader.ReadToEnd();
```

src/shell/dotnet/Shell/App.xaml.cs
ZKRobi: That, or rename the method to something along the lines of ReadPreloadResource

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -175,4 +178,24 @@ private void InjectMessageRouterConfig()
...
+        using (StreamReader reader = new StreamReader(stream))
+        {
+            return reader.ReadToEnd();
```
ZKRobi, line 198: This is picked up by the coverage report that it's not covered, and it sounds like a pretty relevant test for me - if the resource is renamed, or otherwise not loadable, it should fail in CI

src/shell/dotnet/copy-fdc3-iife-bundle.ps1
```diff
@@ -0,0 +1,10 @@
+# Copying the IIFE bundle of FDC3 to the shell' Preload folder
+# so that it can be preloaded for the Embedded Browser
+
+
+$bundle = "./src/shell/js/composeui-fdc3/dist/fdc3-iife-bundle.js"
```
BalassaMarton, line 5: This will only work from the repo root. You can use `$PSScriptRoot` to build the path from the directory where the script is located.

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -46,4 +47,8 @@
 		<ProjectReference Include="..\..\..\messaging\dotnet\src\Host\MorganStanley.ComposeUI.Messaging.Host.csproj" />
 	</ItemGroup>
 
+	<Target Name="PreBuild" BeforeTargets="PreBuildEvent">
+	  <Exec Command="PowerShell cd ..\..\..\..\ ; .\src\shell\dotnet\copy-fdc3-iife-bundle.ps1" />
```
BalassaMarton, line 51: `powershell cd $(ComposeUIRepositoryRoot); .\src\shell\dotnet\copy-fdc3-iife-bundle.ps1`

But it would be better if the script didn't require the repo root as working directory.

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -46,4 +47,8 @@
 		<ProjectReference Include="..\..\..\messaging\dotnet\src\Host\MorganStanley.ComposeUI.Messaging.Host.csproj" />
 	</ItemGroup>
 
+	<Target Name="PreBuild" BeforeTargets="PreBuildEvent">
```
BalassaMarton, line 50: Not sure if this is a good idea. Didn't we already agree on adding the `EmbeddedResource` link directly on the JS output?

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -38,6 +38,7 @@
 	</ItemGroup>
 
 	<ItemGroup>
+		<Folder Include="Preload\" />
```
BalassaMarton, line 41: There's no `EmbeddedResource` item?

kruplm: https://github.com/morganstanley/ComposeUI/commit/7c30c502af23e0f3b578f53471f3ab9cbf3e95f7

Something went wrong with the history...
Let me recommit this

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -14,7 +14,13 @@
...
+	
+	<ItemGroup>
+	  <EmbeddedResource Include="Preload\fdc3-iife-bundle.js">
+	    <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+	  </EmbeddedResource>
+	</ItemGroup>
+	
 	<ItemGroup>
 		<PackageReference Include="Microsoft.Extensions.Hosting" Version="7.0.1" />
 		<PackageReference Include="Microsoft.Web.WebView2" Version="1.0.1462.37" />
```
ztanczos, line 26: Not related to this PR, but do we have dependapot configured for nuget packages? We could use a more recent version of the WebView2 SDK.

bingenito: Yes but I think we can't configure with root so it isn't seeing any of the sln/projects.  We might have to have an entry for each sln

https://github.com/morganstanley/ComposeUI/blob/main/.github/dependabot.yml#L12

kruplm: We had a few dependabot PRs for Nuget last year:
https://github.com/morganstanley/ComposeUI/pull/110
But this project might have been in the root at that time and that's why it was picked up.
Let me revisit the config in a separate PR.

src/shell/dotnet/Shell/Utilities/ResourceReader.cs
bingenito, line 1: New files require license headers

src/shell/dotnet/Shell/Utilities/ResourceReader.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Shell.Utilities
+{
+    public static class ResourceReader
+    {
+        public static string ReadResource(string name) 
+        {
+            var assembly = Assembly.GetExecutingAssembly();
+            string resourcePath = name;
+
+            var stream = assembly.GetManifestResourceStream(resourcePath);
```
ZKRobi, line 32: Streams are disposable, a using statement is in order.

src/shell/dotnet/Shell/Utilities/ResourceReader.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            if (stream == null)
+            {
+                throw new InvalidOperationException("Resource not found");
+            }
+            using (var reader = new StreamReader(stream))
```
ZKRobi, line 38: It's nicer to make use of the new using declaration instead of a using block when the block ends at the end of the parent scope.
e.g. using var reader = new StreamReader(stream)

src/shell/dotnet/tests/Shell.Tests/ResourceReaderTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Shell.Preload;
+
+namespace ShellTests
+{
+    public class ReadResourceTests
+    {
+        [Fact]
+        public void ResourceCanBeRead()
```
ZKRobi, line 30: This test is much more specific, it checks that the Fdc3 bundle is in place, I think the name should also reflect that as that's really what we are interested in.

ZKRobi (changes requested):

BalassaMarton (approved):

PR #326, opened 2023-08-17
Updated Chart and DataGrid examples to use FDC3 standard

lilla28:
Updated example projects to use FDC3 standard.
image

### DIFF
--- a/examples/js-datagrid/src/app/main-view/services/mock-data.service.ts
+++ b/examples/js-datagrid/src/app/main-view/services/mock-data.service.ts
@@ -1,10 +1,9 @@
 /* Morgan Stanley makes this available to you under the Apache License, Version 2.0 (the "License"). You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0. See the NOTICE file distributed with this work for additional information regarding copyright ownership. Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License. */
 
-import { Inject, Injectable } from '@angular/core';
+import { Injectable } from '@angular/core';
 import { interval, Subject } from 'rxjs';
 import { Market } from './mock-data';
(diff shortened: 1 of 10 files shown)


package-lock.json
```diff
@@ -105,7 +105,7 @@
         "@angular/platform-browser": "^15.1.0",
         "@angular/platform-browser-dynamic": "^15.1.0",
         "@angular/router": "^15.1.0",
-        "@morgan-stanley/composeui-messaging-client": "*",
+        "@finos/fdc3": "2.0",
```
kruplm, line 108: That's interesting. The npm package has patch versions, like you use it on the chart.
https://www.npmjs.com/package/@finos/fdc3
I don't know how it hasn't failed :D 
But just to be sure please include the patch version for 2.0.3

src/shell/js/composeui-fdc3/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.1.0",
   "description": "FDC3 DesktopAgent implementation for Compose UI",
   "type": "module",
-  "main": "dist/index.js",
-  "module": "dist/index.js",
+  "main": "dist/esm/index.js",
```
kruplm, line 6: Please revert this. I don't export ESM from this package.

kruplm (changes requested):

package-lock.json
lilla28: Sure will replace it with the patched version!

kruplm (dismissed):

examples/js-datagrid/src/app/main-view/services/mock-data.service.ts
```diff
@@ -13,61 +12,59 @@ export class MockDataService{
...
 
-  private async checkMessageRouterConnection(){
-    if(!this.connected){
-      await this.client.connect();
-      this.connected = true;
+      await window.fdc3.broadcast(context);
     }
   }
+
+  // public async requestSymbolData(topic: string) : Promise<void>{
```
ZKRobi, line 64: Don't leave commented code

lilla28: that was on purpose as per It should (in a later state) involve DataService.
I'll remove it then!

ZKRobi (approved):

PR #329, opened 2023-08-22
Added msbuild target to embed and build composeui-fdc3 package to the Shell project, changed project's files to use filescoped namespace, little refactoring around naming variables

lilla28:
Changes:

- Added msbuild target(s) for building and embed fdc3 bundle. 
Be aware now the bundle file saved into the Output path (bin/Debug(Release)/net6.0-windows) 
---> if we want to include it inside the project directory, I can change it.

- block scoped namespace to file scoped namespace using (to be more consistent around the Shell project)
- Refactoring around the names of the variables (readonly, using var, private, public fields etc.)


src/shell/dotnet/Shell/Utilities/ResourceReader.cs
```diff
@@ -13,28 +13,21 @@
...
+        var assembly = Assembly.GetExecutingAssembly();
 
-            using var stream = assembly.GetManifestResourceStream(resourcePath) ?? throw new InvalidOperationException("Resource not found");
-            using var reader = new StreamReader(stream);
-            {
-                return reader.ReadToEnd();
-            }
+        using var stream = assembly.GetManifestResourceStream(resourcePath) ?? throw new InvalidOperationException("Resource not found");
+        using var reader = new StreamReader(stream);
+        {
```
ztanczos, line 29: Nitpick: { } are unnecessary here

ztanczos:
The bundle is an EmbeddedResource, right? In that case I don't think it should be copied to the output directory as a file, that seems redundant.

src/shell/dotnet/Shell/Utilities/ResourceReader.cs
lilla28: It is a merged code, I didn't want to change that much :) but then I'll remove the {}

src/shell/dotnet/Shell/Fdc3/PreloadFdc3.cs
```diff
@@ -1,13 +1,15 @@
...
 {
-    public static class PreloadFdc3
+    public static readonly string Fdc3BundleResourceName = $@"Shell.bin.{GetEnvironment()}.net6._0_windows.fdc3-iife-bundle.js";
+
+    private static bool _isDebugMode = false;
+    private static string GetEnvironment()
     {
-        public static readonly string Fdc3BundleResourceName = @"Shell.fdc3-iife-bundle.js";
+#if DEBUG
+        _isDebugMode = true;
```
BalassaMarton, line 11: Get methods should not have side effects. `_isDebugMode` is not used elsewhere, you can just put the return into conditional compilation blocks:
```cs
#if DEBUG
        return "Debug";
#else
        return "Release";
#endif
```

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -46,17 +46,37 @@
 		<ProjectReference Include="..\..\..\messaging\dotnet\src\Host\MorganStanley.ComposeUI.Messaging.Host.csproj" />
 	</ItemGroup>
 
-	<PropertyGroup>
-		<Fdc3BundleOutput>$(ComposeUIRepositoryRoot)/src/shell/js/composeui-fdc3/dist/fdc3-iife-bundle.js</Fdc3BundleOutput>
-	</PropertyGroup>
+	<Target Name="InstallMonoRepoDependencies"
+		Condition="!Exists('$(ComposeUIRepositoryRoot)\node_modules')">
+		<Message Importance="high" Text="Installing node_modules..." />
+		<Exec Command="cd $(ComposeUIRepositoryRoot) | npm install" />
```
BalassaMarton, line 52: Are you sure about the pipe characters (here and in other scripts in this file)? That means redirect output. If you mean "do A, and if successful then B" then use `&&`:
```
cd $(ComposeUIRepositoryRoot) && npm install
```

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -46,17 +46,37 @@
...
-		<Exec Command="npx lerna run build --scope @morgan-stanley/composeui-fdc3" />
+	<Target
+		Name="EmbedComposeUIFdc3Bundle"
+		BeforeTargets="PrebuildEvent">
+		<CallTarget Targets="InstallMonoRepoDependencies" />
+		<CallTarget Targets="RunFdc3LernaBuild" />
+		<ItemGroup>
+			<Fdc3BundleOutput Include="$(ComposeUIRepositoryRoot)\src\shell\js\composeui-fdc3\dist\fdc3-iife-bundle.js" />
+		</ItemGroup>
+		<Copy SourceFiles="@(Fdc3BundleOutput)" DestinationFolder="$(TargetDir)\%(RecursiveDir)" />
```
BalassaMarton, line 78: You don't need to copy anything, just use the `Fdc3BundleOutput` items as the `Include` attribute for `EmbeddedResource`.

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -46,17 +46,37 @@
...
+		<Message Importance="high" Text="Running npx lerna run build for composeui-fdc3..." />
+		<Exec Command="cd $(ComposeUIRepositoryRoot) | npx lerna run build --scope @morgan-stanley/composeui-fdc3" />
+		<Message Importance="normal" Text="Finished building the fdc3 bundle." />
+	</Target>
+
+	<Target Name="EmbedFdc3Bundle"
+			Condition="Exists('$(TargetDir)\fdc3-iife-bundle.js')">
+		<Message Importance="normal" Text="Embedding fdc3 resource" />
+		<ItemGroup>
+			<EmbeddedResource Include="$(TargetDir)\%(RecursiveDir)\fdc3-iife-bundle.js" />
```
BalassaMarton, line 66: You're not actually using the `Fdc3BundleOutput` items. Instead, try this:
```
<EmbeddedResource Include="@(Fdc3BundleOutput)" />
```

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -46,17 +46,37 @@
...
+		<Message Importance="normal" Text="Finished command: npm install." />
+	</Target>
 
-	<ItemGroup>
-		<EmbeddedResource Include="$(Fdc3BundleOutput)">
-		</EmbeddedResource>
-	</ItemGroup>
+	<Target Name="RunFdc3LernaBuild">
+		<Message Importance="high" Text="Running npx lerna run build for composeui-fdc3..." />
+		<Exec Command="cd $(ComposeUIRepositoryRoot) | npx lerna run build --scope @morgan-stanley/composeui-fdc3" />
```
BalassaMarton, line 58: Suspicious piping instead of `&&`

BalassaMarton (changes requested):

src/shell/dotnet/Shell/Fdc3/PreloadFdc3.cs
lilla28: removed, as it is unnecessary

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -47,16 +47,31 @@
...
+		<Message Importance="normal" Text="Finished command: npm install." />
 	</Target>
+
+	<Target Name="RunFdc3LernaBuild">
+		<Message Importance="high" Text="Running npx lerna run build for composeui-fdc3..." />
+		<Exec Command="cd $(ComposeUIRepositoryRoot) &amp; npx lerna run build --scope @morgan-stanley/composeui-fdc3" />
+		<Message Importance="normal" Text="Finished building the fdc3 bundle." />
+	</Target>
+
+	<Target
```
BalassaMarton, line 70: I'd add back the condition, otherwise this will run with every build, even if the bundle exists. 
Maybe even remove the separate RunFdc3LernaBuild target, as it is not triggered from anywhere else.
```xml
    <Target Name="BuildComposeUIFdc3Bundle" BeforeTargets="PrebuildEvent">        
        <CallTarget Targets="InstallMonoRepoDependencies" />
        <Message Importance="high" Text="Running npx lerna run build for composeui-fdc3..." />
        <Exec Command="cd $(ComposeUIRepositoryRoot) &amp;&amp; npx lerna run build --scope @morgan-stanley/composeui-fdc3" />
        <Message Importance="normal" Text="Finished building the fdc3 bundle." />
    </Target>
```

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -47,16 +47,31 @@
...
+		Name="BuildComposeUIFdc3Bundle"
+		BeforeTargets="PrebuildEvent">
+		<CallTarget Targets="InstallMonoRepoDependencies" />
```
lilla28, line 73: This won't be called if the node_modules folder exists, but the other target will be executed.
The Choose and When and OtherWise cannot be used within a Target.

BalassaMarton (changes requested):

src/shell/dotnet/Shell/Shell.csproj
BalassaMarton: I managed to leave the condition out of the target...
```
    <Target Name="BuildComposeUIFdc3Bundle" BeforeTargets="PrebuildEvent" Condition="!Exists('$(Fdc3BundleOutput)')">        
        <CallTarget Targets="InstallMonoRepoDependencies" />
        <Message Importance="high" Text="Running npx lerna run build for composeui-fdc3..." />
        <Exec Command="cd $(ComposeUIRepositoryRoot) &amp;&amp; npx lerna run build --scope @morgan-stanley/composeui-fdc3" />
        <Message Importance="normal" Text="Finished building the fdc3 bundle." />
    </Target>
```

PR #347, opened 2023-09-01
Fdc3 - Enable Fdc3 as an optional feature

lilla28:
Changes:
- Added `Directory.Build.props` files
- Moved `composeui-fdc3` js implementation under `fdc3/js` folder
- Renamed `PreloadFdc3` to `ResourceNames`
- Added `MessagingRouter.Client` to the solution for building the project
- Refactored namespaces in `Shell` and `Shell.Tests` projects, 
- Added `Fdc3Options` to check if it is enabled from `appsettings.json` 
- Renamed `Fdc3DesktopAgentConfig` to `Fdc3DesktopAgentBuilder`
- Modified `ServiceCollectionExtensions` in `Fdc3DesktopAgent `
- Added `ServiceCollection` extension method for `DesktopAgent` to check the `Fdc3Options` and inject `DesktopAgent` if that is enabled from config
- Added TODO to `ResourceReader` that we should move this utility to a new project later under `src/shared/dotnet`
- Added missing headers
- Fixing `DesktopAgent.Tests`'s `DesktopAgent` setup 
- Removed `WithUserChannel` from `DesktopAgent`'s `Fdc3DesktopAgentBuilder`
- Removed internal `BuilderActions` property from `Fdc3Options`


src/fdc3/Directory.Build.props
```diff
@@ -0,0 +1,3 @@
+<Project>
+	<Import Project="../Directory.Build.props" Condition="Exists('../Directory.Build.props')" />
```
BalassaMarton, line 2: Don't remove now that they are there, but these files (with the `Import` for the parent directory) are only necessary when they have content, otherwise the `Directory.Build.props` file closest to the project is imported automatically.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Utilities/ResourceReader.cs
```diff
@@ -12,11 +12,9 @@
...
  */
 
-using System;
-using System.IO;
 using System.Reflection;
 
-namespace Shell.Utilities;
+namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Utilities;
 
 public static class ResourceReader
```
BalassaMarton, line 19: Later we should move this utility to a new project under `src/shared/dotnet`

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Common/PreloadFdc3Names.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Reflection;
+
+namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Common;
+
+public static class PreloadFdc3Names
```
BalassaMarton, line 5: 'Preload' is something that only makes sense in the context of the embedded web browser. Since these are resource names, you can just call this class `ResourceNames`, and its members accordingly, like `Fdc3Bundle`.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/DesktopAgent.csproj
```diff
@@ -1,21 +1,41 @@
...
 
-  <ItemGroup>
-    <ProjectReference Include="..\..\..\..\..\messaging\dotnet\src\Core\MorganStanley.ComposeUI.Messaging.Core.csproj" />
-  </ItemGroup>
+	<ItemGroup>
+		<ProjectReference Include="..\..\..\..\..\messaging\dotnet\src\Core\MorganStanley.ComposeUI.Messaging.Core.csproj" />
+	</ItemGroup>
+
+	<PropertyGroup>
+		<Fdc3BundleOutput>$(ComposeUIRepositoryRoot)\src\fdc3\js\composeui-fdc3\dist\fdc3-iife-bundle.js</Fdc3BundleOutput>
```
BalassaMarton, line 22: I'm not sure it's a good idea to move the embedded JS bundle to the Fdc3 dotnet project, it doesn't belong there. The shell is what wires these together, but the desktop agent implementation for .NET doesn't depend on the JS package. I'd move the resource back to the shell project and make it optional to 
- add the FDC3 backend
- add the JS bundle to preload scripts

src/fdc3/Directory.Build.props
lilla28: Understood. The reason why I have added to this folder is because the DesktopAgent project had the Directory.Build.props file before, and it was searching for the props file above the folder, and that's why It couldn't resolve the common properties when I added the targets for the project.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/DependencyInjection/Fdc3DesktopAgentBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+public class Fdc3DesktopAgentBuilder
+{
+    public IServiceCollection ServiceCollection { get; }
+
+    public Fdc3DesktopAgentBuilder(IServiceCollection serviceCollection)
+    {
+        ServiceCollection = serviceCollection;
+    }
+
+    public Fdc3DesktopAgentBuilder WithUserChannel(string channelId)
```
BalassaMarton, line 28: Now that the user channel can be set in options, this method can be removed. `Fdc3DesktopAgent.StartAsync` should directly call `AddUserChannel` if the `ChannelId` is set in options (in other words, setting the channel shouldn't be a builder action any more).

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -27,14 +28,17 @@ public class Fdc3DesktopAgent : IHostedService
...
     private readonly ILoggerFactory _loggerFactory;
+    private readonly Fdc3Options _options;
     private readonly IMessageRouter _messageRouter;
-    private Fdc3DesktopAgentConfig? _config;
 
-    public Fdc3DesktopAgent(IMessageRouter messageRouter, ILoggerFactory? loggerFactory, Fdc3DesktopAgentConfig? config)
+    public Fdc3DesktopAgent(
+        IOptions<Fdc3Options> options, 
+        IMessageRouter messageRouter, 
+        ILoggerFactory? loggerFactory)
```
BalassaMarton, line 37: Add default value (`ILoggerFactory? loggerFactory = null`) so that it can be optional in DI. Otherwise you'll get an exception instead of `null`.

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -113,8 +112,20 @@ private void ConfigureServices(HostBuilderContext context, IServiceCollection se
...
+        //This can be replaced by the `InjectFdc3BackendServiceIfEnabledFromConfig` extension method
+        ///*
+        var fdc3Configuration = context.Configuration
+            .GetSection(Fdc3Options.Fdc3OptionsName);
+
+        var fdc3Options = fdc3Configuration
+            .Get<Fdc3Options>();
 
-        services.AddFdc3DesktopAgent(fdc3 => fdc3.WithUserChannel("default"));
+        if (fdc3Options != null && fdc3Options.EnableFdc3)
```
BalassaMarton, line 123: Suggested TODO item: Instead of having EnableFdc3 inside Fdc3Options, this should be a feature toggle once we have feature toggles.

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -113,8 +112,20 @@ private void ConfigureServices(HostBuilderContext context, IServiceCollection se
...
+        {
+            services.Configure<Fdc3Options>(fdc3Configuration);
+            services.AddFdc3DesktopAgent(builder => builder.WithUserChannel(fdc3Options.ChannelId ?? "default"));
```
BalassaMarton, line 126: I'd remove `WithUserChannel` and just add a `Configure<Fdc3Options>` step that applies this default.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -22,15 +23,45 @@ public static class ServiceCollectionExtensions
...
+    public static IServiceCollection InjectFdc3BackendServiceIfEnabledFromConfig(
+        this IServiceCollection serviceCollection,
+        IConfiguration configuration,
+        Action<Fdc3DesktopAgentBuilder>? builderAction = null)
+    {
+        var fdc3Options = new Fdc3Options();
+
+        var fdc3Configuration = configuration.GetSection(Fdc3Options.Fdc3OptionsName);
+
+        fdc3Configuration.Bind(fdc3Options);
```
ztanczos, line 57: Nitpick: would it be more convenient to simply use ConfigurationBinder.Get<T>?

lilla28: Sure, I will replace it, but the only reason why I haven't used it, because right now the whole configuration is passed to the method, which will look for the `Fdc3Options` IConfigurationSection. And with this solution (with the `ConfigurationBinder`) the user only could pass that section, otherwise - if the user will pass the whole config it will return the default value. I added a descript for this parameter, now that i replaced with `ConfigurationBinder` solution :)

ztanczos: I mean you can use `configuration.GetSection(...).Get<T>()`

src/shell/dotnet/Shell/Shell.csproj
```diff
@@ -7,7 +7,8 @@
 		<UseWPF>true</UseWPF>
 		<EnableDefaultPageItems>false</EnableDefaultPageItems>
 		<LangVersion>preview</LangVersion>
-		<AssemblyName>ComposeUI-Shell</AssemblyName>
+		<AssemblyName>MorganStanley.ComposeUI.Shell</AssemblyName>
+		<RootNamespace>MorganStanley.ComposeUI.$(MSBuildProjectName.Replace(" ", "_"))</RootNamespace>
```
BalassaMarton, line 11: How about `<RootNamespace>MorganStanley.ComposeUI.Shell</RootNamespace>`? We might rename the project file, but we probably don't want the namespace to change as well.

src/shell/dotnet/tests/Shell.Tests/Shell.Tests.csproj
```diff
@@ -1,11 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net6.0-windows</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-
     <IsPackable>false</IsPackable>
+	<RootNamespace>MorganStanley.ComposeUI.$(MSBuildProjectName.Replace(" ", "_"))</RootNamespace>
```
BalassaMarton, line 8: `<RootNamespace>MorganStanley.ComposeUI.Shell</RootNamespace>`

src/shell/dotnet/add-to-path.ps1
```diff
@@ -9,5 +9,6 @@ if ($paths.indexof($shellPath.Path) -ne -1) { exit }
 
 $paths += $shellPath.Path
 $path = $paths -join ";"
+echo $path;
```
BalassaMarton, line 12: Do you want to keep this?

BalassaMarton (dismissed):

ZKRobi (approved):

kruplm (approved):

PR #377, opened 2023-10-16
Initial Module Loader implementation

ztanczos:
- Refactored & added Module Loader to src
- Module Loader is available via DI by calling `AddModuleLoader()`
- Added unit tests
- Stopping modules is not implemented yet
- NativeModuleRunner is not implemented yet


src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader/ModuleLoader.cs
```diff
@@ -0,0 +1,106 @@
...
+        Guid instanceId = Guid.NewGuid();
+        var moduleInstance = new ModuleInstance(instanceId, manifest, request);
+        _modules.TryAdd(instanceId, moduleInstance);
+
+        _lifetimeEvents.OnNext(new LifetimeEvent.Starting(moduleInstance));
+        var startupContext = new StartupContext(request);
+
+        int index = -1;
+        Func<Task> nextAction = null!;
+        nextAction = () =>
```
BalassaMarton, line 64: This is a bit hazardous, if I call the same `next` multiple times from my startup action, I expect the same (next) startup action to be called. Maybe a unique delegate should be created for each startup action.
```cs
var pipeline = _startupActions
    .Reverse()
    .Aggregate(
        () => Task.CompletedTask, 
        (next, action) => () => action.InvokeAsync(startupContext, next));

await moduleRunner.Start(moduleInstance, startupContext, pipeline);
```

src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader.Abstractions/IModuleRunner.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace MorganStanley.ComposeUI.ModuleLoader;
+
+public interface IModuleRunner
+{
+    string ModuleType { get; }
+
+    Task Start(IModuleInstance moduleInstance, StartupContext startupContext, Func<Task> pipeline);
+
+    Task Stop(IModuleInstance moduleInstance, CancellationToken cancellationToken = default);
```
BalassaMarton, line 21: Do we need cancellation here? I'm not sure it's feasible to cancel, other than from the module itself (we'll need some mechanism to do that, but it's not by a cancellation token)

ztanczos: I don't know if we need cancellation here but I agree we don't need cancellation here *right now* :) I'll remove that argument.

src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader/ModuleLoader.cs
ztanczos: Nice tip, thank you. Updated the code

src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader.Abstractions/IModuleRunner.cs
BalassaMarton: It's just that if you cancel it halfway, what's the expected behavior?

src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader/ModuleLoader.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            return;
+        }
+
+        await StopModuleInternal(module);
+    }
+
+    public async ValueTask DisposeAsync()
+    {
+        _lifetimeEvents.Dispose();
```
BalassaMarton, line 87: Move this to the end of the method? `StopModuleInternal` will still fire the lifetime events and maybe that's a good thing (we can't really make assumptions about when `DisposeAsync` is called)

BalassaMarton (approved):

src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader/ModuleLoader.cs
ztanczos: Let's do this as a follow-up

ZKRobi (approved):

PR #382, opened 2023-10-19
Integrate Module Loader into the shell

BalassaMarton:
- Added the Module Loader to the Shell
- Implemented a simple Module Catalog (supporting only web modules)
- WebWindow now loads the preload scripts dynamically using script provider abstraction
- Moved startup logic to startup actions
- Revamped main window
- Removed the example manifest and added a new one to a dedicated examples folder


src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader/ModuleLoader.cs
```diff
@@ -47,7 +47,7 @@ public async Task<IModuleInstance> StartModule(StartRequest request)
             throw new Exception($"Unknown Module id: {request.ModuleId}");
         }
 
-        if (!_moduleRunners.TryGetValue(request.ModuleId, out var moduleRunner))
+        if (!_moduleRunners.TryGetValue(manifest.ModuleType, out var moduleRunner))
```
ztanczos, line 50: Good catch, ty 👍

ztanczos (approved):

PR #383, opened 2023-10-27
feat(fdc3): Implement RaiseIntent, AddIntentListener, FindIntent and FindIntentsByContext

lilla28:
Implemented: 

-  RaiseIntent functionality: we register the service to the specific topic and we are listening for the raiseIntent calls from the clients. We are checking if the necessary information is contained by the RaiseIntentRequest message, and after that we are selecting the appropriate Fdc3Apps from both from the running applications (tracked by the ModuleLoader) and from the AppDirectory. To keep the consistency of the raiseIntent, findIntent, and findIntentsByContext we will send back AppIntent type as RaiseIntentResponse, if we found any acceptable app. If multiple AppIntents found then we will send back an error message.
If the AppIntents conatins more than 1 app we will register a service for the originating app to listen for its choice (selected app from the ResolverUI). If it contains just one app, then we will check if it is already running or it needs to be stated to handle the intent. If it needs to be started then we will save the context to a Dictionary, so when the ModuleLoader appropriately started the app, we will pass the context its IntentListener. If the app does not need to be started then we will just publish the Context to its IntentListener.

- AddIntentListener proposed functionality: we will subscribe to an appropriate topic (raiseIntent/intent/instanceIdOfHandlingApp), and when an appropriate type of message comes, then the client should handle with its intentHandler, and invoke a registered service on the backend (StoreIntentResult). When it is called it will save the IntentResult to a collection containing the originating app instanceId, and the target instanceId (what called the raiseIntent). If the IntentResolution (result of the raiseIntent) will call its getResult() method, then we will retrieve the IntentResult from this collection.

- FindIntent funcionality: searches via the backend for the appropriate apps from both of the running apps and from the AppDirectory with the appropriate query. Returns AppIntent, if multiple AppIntent found, it will retrieve an error message.

- FindIntentsByContext funcionality: searches via the backend for the appropriate apps from both of the running apps and from the AppDirectory with the appropriate query. Returns an AppIntent


src/fdc3/dotnet/DesktopAgent/tests/DesktopAgent.Tests/TestUtils/MockAppDirectory.cs
```diff
@@ -0,0 +1,263 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+using MorganStanley.Fdc3.AppDirectory;
+
+namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Tests.TestUtils;
+
+public class MockAppDirectory : IAppDirectory
```
BalassaMarton, line 19: Don't mock types that are not owned by the current project (in this case, the solution). IAppDirectory already has an implementation that is tested and can load itself from JSON.

BalassaMarton (changes requested):

src/shell/dotnet/Shell/Fdc3/Fdc3StartupAction.cs
```diff
@@ -23,8 +23,22 @@ public Task InvokeAsync(StartupContext startupContext, Func<Task> next)
     {
         if (startupContext.ModuleInstance.Manifest.ModuleType == ModuleType.Web)
         {
-            startupContext.GetOrAddProperty<WebStartupProperties>()
+            var webProperties = startupContext.GetOrAddProperty<WebStartupProperties>();
+            webProperties
                 .ScriptProviders.Add(_ => new ValueTask<string>(ResourceReader.ReadResource(ResourceNames.Fdc3Bundle)));
+
+            webProperties
```
BalassaMarton, line 30: This should be either added by the shell (because it's not FDC3-specific), or added under a different name, eg. `window.composeui.fdc3.instanceConfig`

prototypes/process-explorer/js/.gitignore
```diff
@@ -0,0 +1,42 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
```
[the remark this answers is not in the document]
lilla28: when I ran the lerna command it always created a bunch of output files for this project and there was a time when I pushed my code with the output and I needed to clenup, thats why I have added the .gitignore files to this project.

prototypes/process-explorer/.gitignore
```diff
@@ -0,0 +1,42 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
```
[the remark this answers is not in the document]
lilla28: see above

BalassaMarton (changes requested):

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Contracts/RaiseIntentResponse.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        var raiseIntentResponse = new RaiseIntentResponse()
+        {
+            Intent = intent
+        };
+
+        if (appMetadatas == null) return raiseIntentResponse;
+        
+        //Could not handle the IAppMetadata interface when deserializing/serializing, that's why we would need a casting. 
+        var castedAppMetadata = appMetadatas.Select(
```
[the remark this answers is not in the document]
lilla28: So as I was handling this feedback, I realized why I did this. So for consistency I was giving back `IAppIntent` (which can be `AppIntent` ofc), but for `RaiseIntentResponse` I would need just the `AppMetadata` from the `AppIntent`.
Now the `AppIntent`'s `Apps` property returns `IEnumerable<IAppMetadata>` not `IEnumerable<AppMetadata>` so we either cast here or in the response message (or define other possibilities). 
mentioned code snippet: https://github.com/morganstanley/ComposeUI/pull/383/files#diff-d63b5beccf11399401fbd494c7a34834af23bc11dc154a1bcdc5486663c2b1dcR181-R195

BalassaMarton (changes requested):

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Converters/IAppMetadataJsonConverter.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public override IAppMetadata? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        return JsonSerializer.Deserialize<AppMetadata>(ref reader, options);
+    }
+
+    public override void Write(Utf8JsonWriter writer, IAppMetadata value, JsonSerializerOptions options)
+    {
+        JsonSerializer.Serialize(writer, (AppMetadata)value, options);
+        var test = JsonSerializer.Serialize((AppMetadata) value, options);
```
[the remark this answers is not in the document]
lilla28: They are used, when serializing the AppIntent: <seealso cref="https://github.com/morganstanley/fdc3-dotnet/blob/main/src/Fdc3/AppIntent.cs#L31-L33">

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -73,19 +215,618 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
+        return invocation.RaiseIntentMessageId;
+    }
+
+    //Publishing intent resolution request to the fdc3 clients, they will receive the message and start their intenthandler appropriately, and send a store request back to the backend.
+    private async Task RaiseIntentResolution(string raisedIntentMessageId, string intent, Context context, string targetId, string sourceFdc3InstanceId)
+    {
+        if (_runningApplications.TryGetValue(new(sourceFdc3InstanceId), out var sourceApp))
+        {
+            var sourceAppIdentifier = new AppIdentifier(sourceApp.AppId, sourceFdc3InstanceId);
+            var intentResolutionRequest = MessageBuffer.Factory.CreateJson(new RaiseIntentResolutionRequest(raisedIntentMessageId, context, new ContextMetadata(sourceAppIdentifier)), _jsonSerializerOptions);
```
ZKRobi, line 535: Adding the app id into the metadata seems odd as the message is not originating from the app. Am I missing something?

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
ZKRobi, line 1: Omitting braces around one-liner conditional operations is not a good practice, it quite often results in bugs e.g. when someone else adds another line, forgetting to add the braces.

ZKRobi (changes requested):

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -48,22 +84,128 @@ public async ValueTask AddUserChannel(string id)
         _userChannels.Add(uc);
     }
 
-    internal async ValueTask<MessageBuffer?> FindChannel(string _, MessageBuffer? requestBuffer, MessageContext _2)
+    internal ValueTask<MessageBuffer?> FindChannel(string endpoint, MessageBuffer? payload, MessageContext context)
```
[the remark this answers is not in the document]
lilla28: FindChannel was implemented by you.  Previously I was told to suppress the warnings around this method(lacks of await), and maybe I have just done some refactoring around the result to use the ? : operator. I can change back to normal conditional like it was if that is more understandable.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
lilla28: so the whole logic looks like that If an app raises an intent to a literally existing and running app, then the solver app would get the AppIdentifier from the originating one while resolving the intent.
The line above queries the running applications (fdc3 instance ids - FDC3App type dictionary) and sends the ContextMetadata to the target app to solve it with their IntentHandler(context, contextMetadata?).

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
lilla28: I like this style, but okay I'll modify it :)

src/shared/dotnet/src/MorganStanley.ComposeUI.Utilities/ResourceNames.cs
```diff
@@ -14,10 +14,10 @@
 
 using System.Reflection;
 
-namespace MorganStanley.ComposeUI.Shell.Utilities;
+namespace MorganStanley.ComposeUI.Utilities;
 
 public static class ResourceNames
 {
-    public static readonly string Fdc3Bundle = @$"{Assembly.GetExecutingAssembly().ManifestModule.Assembly.GetName()?.Name}.fdc3-iife-bundle.js";
+    public static readonly string Fdc3Bundle = @$"{Assembly.GetCallingAssembly().ManifestModule.Assembly.GetName()?.Name}.fdc3-iife-bundle.js";
```
BalassaMarton, line 21: These resource names are project specific, move them accordingly

ZKRobi (changes requested):
I added quite a few comments, most of them are typos or other suggestions for comments and naming.
There are a few that we should discuss as a follow-up, and should not be added as scope to this PR. I added those comments to support discussion.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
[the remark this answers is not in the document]
ZKRobi: Well, I'm caught guilty I guess :)

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/IFdc3DesktopAgentBridge.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <param name="request"></param>
+    /// <returns></returns>
+    public ValueTask<GetIntentResultResponse> GetIntentResult(GetIntentResultRequest request);
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="request"></param>
+    /// <returns></returns>
+    public ValueTask<KeyValuePair<RaiseIntentResponse, RaiseIntentResolutionMessage?>> RaiseIntent(RaiseIntentRequest request);
```
BalassaMarton, line 84: How about returning a dedicated type or a tuple here and in similar methods?

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Protocol/AppIntent.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using MorganStanley.Fdc3;
+
+namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Protocol;
+
+public class AppIntent : IAppIntent
+{
+    public IIntentMetadata Intent { get; set; }
+
+    public IEnumerable<IAppMetadata> Apps { get; set; }
```
BalassaMarton, line 21: I'd rather use the implementation type in these properties with the interface property implemented explicitly, eg.

```cs
public IEnumerable<AppMetadata> Apps { get; set; }
IEnumerable<IAppMetadata> IAppIntent.Apps => Apps;
```

Then you can get rid of the explicit casts needed when converting to/from `IEnumerable<IAppMetadata>` etc.
This applies to the other types in this namespace as well.

src/fdc3/dotnet/DesktopAgent/tests/DesktopAgent.Tests/Helpers/Fdc3InstanceIdRetriever.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Tests.Helpers;
+
+internal static class Fdc3InstanceIdRetriever
+{
+    public static string Get(IModuleInstance instance)
+    {
+        var fdc3InstanceId = instance.StartRequest.Parameters.FirstOrDefault(parameter => parameter.Key == Fdc3StartupParameters.Fdc3InstanceId);
+        if (string.IsNullOrEmpty(fdc3InstanceId.Value))
+            if (instance.GetProperties().FirstOrDefault(property => property is Fdc3StartupProperties) is not Fdc3StartupProperties fdc3StartupProperties)
```
BalassaMarton, line 26: Add braces please

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/DesktopAgent.csproj
```diff
@@ -3,6 +3,7 @@
 		<TargetFramework>net6.0</TargetFramework>
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
+		<LangVersion>preview</LangVersion>
```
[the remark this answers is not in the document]
lilla28: should be :)
as Fdc3StartupAction's copied implementation is using that for interpolated raw string literal

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Infrastructure/Fdc3DesktopAgentMessageRouterService.cs
```diff
@@ -0,0 +1,301 @@
...
+        {
+            return MessageBuffer.Factory.CreateJson(response.Key, _jsonSerializerOptions);
+        }
+
+        if (response.Value != null && response.Value.Any())
+        {
+            foreach (var message in response.Value)
+            {
+                await _messageRouter.PublishAsync(
+                    Fdc3Topic.RaiseIntentResolution(message.Intent, message.TargetModuleInstanceId),
```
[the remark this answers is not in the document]
lilla28: Yes, the typescript client when it adds its listener.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -12,80 +12,799 @@
...
+                return new(RaiseIntentResponse.Success(raisedIntentMessageId, intent, targetAppMetadata), resolution);
+            }
+            else
+            {
+                try
+                {
+                    var fdc3InstanceId = Guid.NewGuid();
+                    var moduleInstance = await _moduleLoader.StartModule(
+                        new StartRequest(
+                            targetAppMetadata.AppId, //TODO: possible remove some identifier like @"fdc3."
```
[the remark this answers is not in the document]
lilla28: yes previously we talked about that we will somehow differentiate the modulecatalog and the appdirectory with some prefix or something, those todos just reminders.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -12,80 +12,799 @@
...
+                }
+            }
+        }
+    }
+
+    private async Task<RaiseIntentResolutionInvocation?> IsIntentResolved(GetIntentResultRequest request)
+    {
+        RaiseIntentResolutionInvocation? resolution = null;
+        while (
+                !_raisedIntentResolutions.TryGetValue(new Guid(request.TargetAppIdentifier.InstanceId!), out var resolver)
```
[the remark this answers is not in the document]
lilla28: As previously was sugegsted to remove the lock for this concurrentdictionary, this step is no longer necessary I assume

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/UserChannel.cs
```diff
@@ -53,27 +50,34 @@ public async ValueTask Connect()
...
+        await _messageRouter.ConnectAsync();
+
+        var broadcastObserver = AsyncObserver.Create<TopicMessage>(x => HandleBroadcast(x.Payload));
 
-        var broadcastSubscribing = _desktopAgentService.SubscribeAsync(_topics.Broadcast, HandleBroadcast, CancellationToken.None);
+        var broadcastSubscribing = _messageRouter.SubscribeAsync(_topics.Broadcast, broadcastObserver);
 
-        await _desktopAgentService.RegisterServiceAsync<GetCurrentContextRequest>(_topics.GetCurrentContext, GetCurrentContext);
-        
+        //await _messageRouter.RegisterEndpointAsync(_topics.GetCurrentContext, GetCurrentContext);
```
ZKRobi, line 59: Any reason for leaving this commented instead of releting?

lilla28: This is the original file, I just reverted back that:
https://github.com/morganstanley/ComposeUI/blob/main/src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/UserChannel.cs#L59

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Converters/IAppIntentJsonConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using MorganStanley.Fdc3;
+using AppIntent = MorganStanley.ComposeUI.Fdc3.DesktopAgent.Protocol.AppIntent;
+
+namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Converters;
+
+public class IAppIntentJsonConverter : JsonConverter<IAppIntent>
```
ZKRobi, line 22: class IAppIntent...?

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/UserChannel.cs
ZKRobi: Feel free to fix my mistakes ;)

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Converters/IAppIntentJsonConverter.cs
lilla28: Yeah, I haven't know how to name it I'll rename all of the related classes. :)

src/fdc3/dotnet/DesktopAgent/tests/DesktopAgent.Tests/Fdc3DesktopAgentTests.cs
```diff
@@ -54,8 +53,7 @@ public async Task AddUserChannel_wont_throw()
             mockMessageService.Object,
             NullLogger<UserChannel>.Instance);
 
-        var action = async () => await _fdc3.AddUserChannel(mockUserChannel.Object);
-        await action.Should().NotThrowAsync();
+        new Action(() => _fdc3.AddUserChannel(mockUserChannel.Object)).Should().NotThrow();
```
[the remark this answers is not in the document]
lilla28: Fixed, I will push in the new commit later.

kruplm:
TypeScript code LGTM

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Infrastructure/Internal/Fdc3DesktopAgentMessageRouterService.cs
```diff
@@ -0,0 +1,250 @@
...
+        await _messageRouter.RegisterServiceAsync(Fdc3Topic.FindIntent, HandleFindIntent, cancellationToken: cancellationToken);
+        await _messageRouter.RegisterServiceAsync(Fdc3Topic.FindIntentsByContext, HandleFindIntentsByContext, cancellationToken: cancellationToken);
+        await _messageRouter.RegisterServiceAsync(Fdc3Topic.RaiseIntent, HandleRaiseIntent, cancellationToken: cancellationToken);
+        await _messageRouter.RegisterServiceAsync(Fdc3Topic.GetIntentResult, HandleGetIntentResult, cancellationToken: cancellationToken);
+        await _messageRouter.RegisterServiceAsync(Fdc3Topic.SendIntentResult, HandleStoreIntentResult, cancellationToken: cancellationToken);
+        await _messageRouter.RegisterServiceAsync(Fdc3Topic.AddIntentListener, HandleAddIntentListener, cancellationToken: cancellationToken);
+
+        await _desktopAgent.StartAsync(cancellationToken);
+
+        if (_options.ChannelId != null)
```
BalassaMarton, line 228: Can we move this logic to `Fdc3DesktopAgent`'s `StartAsync` method instead? This behavior is independent of the messaging layer. `HandleAddUserChannel` can be completely removed since it is only called by tests (those should test `Fdc3DesktopAgent` instead) and this method which discards the return value anyway.

BalassaMarton (changes requested):

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Infrastructure/Internal/Fdc3DesktopAgentMessageRouterService.cs
lilla28: So UserChannel is currently implemented with the MessageRouter layer, so I can't pass the IMessageRouter client to the StartAsync method of the Fdc3DesktopAgent class without losing the messaging layer independency. 
It's okay to use the Fdc3DesktopAgentOptions within the Fdc3DesktopAgent class but I can't instantiate a UserChannel class without refactoring that class. (previously I was told to avoid that in this PR.)

BalassaMarton (approved):

PR #415, opened 2024-01-18
feat(examples): Modified Chart and Datagrid examples to use the raiseIntent and addIntentListener methods

lilla28:
Changes:
- Moved js-datagrid and js-chart under a different folder under the examples,
- Updated dependencies for those 2 example projects
- Added raiseIntent and addIntentListener usings 
- Added sample (Angular icon) to the js-datagrid on the Shell's Ribbon menu (can be removed), for that I also used the previously implemented EnvironmentImageSourcePolicy


src/shell/dotnet/examples/module-catalog.json
```diff
@@ -34,5 +34,22 @@
...
         }
-    }
+    },
+	{
+		"Id": "Datagrid Example",
+        "Name": "Js Datagrid Example",
+        "ModuleType": "web",
+        "Details": {
+            "Url": "http://localhost:4200/",
+			"IconUrl": "https://brandslogos.com/wp-content/uploads/images/large/angular-icon-logo.png"
```
kruplm, line 44: Developers would just use the favicon here, similar to what you see on line 35.
I checked the favicon is there for the project so please just set that here.
It's available from the url path like "http://localhost:4200/favicon.ico"

However I like the idea of showcasing "COMPOSE_ALLOWED_IMAGE_SOURCES", but from at the moment what we have available the chart or one of the simple url loads would be better candidates.
If you'd like to do an example for this on this PR I'd suggest to go with the chart and use an icon that's under the creative commons license like e.g: https://commons.wikimedia.org/wiki/File:Simpleicons_Business_bars-chart-ascending.svg.

But it would be better to just create a very simple example that showcases only this feature on a separate PR.

examples/fdc3-appdirectory/apps-with-intents.json
kruplm, line 1: Nitpick: indentation should be fixed.

kruplm (changes requested):

kruplm (approved):

src/shell/dotnet/Shell/MainWindow.xaml.cs
```diff
@@ -51,6 +51,9 @@ private async void RibbonWindow_Initialized(object sender, System.EventArgs e)
         var modules = new List<ModuleViewModel>();
         foreach (var moduleId in moduleIds)
         {
+            // We dont want to see the chart example on the Ribbon menu, but without initializing in the module catalog we can't start via raiseIntent call
```
ztanczos, line 54: Why don't we want to see the chart example on the Ribbon menu?

BalassaMarton (changes requested):

src/shell/dotnet/Shell/MainWindow.xaml.cs
lilla28: Because it has no purpose. Right now the chart add its intent listener on load event. As right now the implementation of raiseIntent/addIntentListener could not popup a resolve UI to pick the right application(running instance), selecting a row from the Grid would just open a new Chart window even though we already opened the Chart from the Ribbon menu:
https://github.com/morganstanley/ComposeUI/blob/main/src/fdc3/js/composeui-fdc3/src/ComposeUIDesktopAgent.ts#L158

It picks up the first item from the response which should contain 2 elements. One is the Chart (not opened), and one is an instance of the chart.
As the first element would be the not running Chart- that can be opened, I do not see the purpose to set the Chart on the Ribbon menu, bacuse it wont do anything. :)

src/shell/dotnet/Shell/MainWindow.xaml.cs
```diff
@@ -51,6 +51,9 @@ private async void RibbonWindow_Initialized(object sender, System.EventArgs e)
...
+            if (moduleId == "Chart Example") continue;
```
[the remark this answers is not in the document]
lilla28: See above, I can add the chart, but still i do not see the purpose of it.
I understand that we can't bake into the Shell. I will figure out something.

examples/fdc3-chart-and-grid/js-datagrid/src/app/main-view/services/mock-data.service.ts
```diff
@@ -12,52 +13,67 @@ export class MockDataService{
...
+      if (symbol) {
+        marketSymbol = this.market.createMarketSymbol(symbol);
+      }
+      
       const context: Context = {
-        type: 'fdc3.instrument',
+        type: ContextTypes.Chart,
         id: {
           ticker: marketSymbol?.symbol,
           buyData: marketSymbol?.buy,
```
ZKRobi, line 40: I know this is not new code, but this buyData and sellData is not part of the standard and in this sense, it's a misuse.
As I see it, the intent we raise should only contain that it's a ViewChart intent, and the instrument info. The rest of the data should be transferred via other means (preferably via a standard contexts for the sake of this example, but I'm not quite sure which one would fit this purpose). This would also add the possibility to "spice up" the example with "live data". We might want to discuss how to follow up on this offline :)

src/shell/dotnet/Shell/MainWindow.xaml.cs
ZKRobi: Just as food for thought, as it really doesn't make much sense to open the Chart without any selected symbol, but for the sake of the example it might make sense to show that it can register an intent handler and listen to it even in that case.

src/shell/dotnet/Shell/MainWindow.xaml.cs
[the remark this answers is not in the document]
ZKRobi: As discussed offline, it is probably better to make the chart available from the ribbon and handle the initial state in the app

examples/fdc3-chart-and-grid/js-chart/chart.js
```diff
@@ -37,16 +37,16 @@ window.addEventListener('load', function() {
...
   (async () => {
-    currentChannel = await window.fdc3.getCurrentChannel();
-  
-    if(!currentChannel) {
-      await window.fdc3.joinUserChannel("default");
-    }
-
-    await window.fdc3.addContextListener('fdc3.instrument', (context, metadata) => {
-      if(context.id) {
+    intentListener = await window.fdc3.addIntentListener("ViewChart", (context, contextMetada) => {
```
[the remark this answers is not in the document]
ZKRobi: As discussed offline, these cases should be handled:
 - The view chart button sends an intent for the relevant instrument
 - Selecting a row sends a context that will be handled if the chart is open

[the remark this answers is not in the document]
ZKRobi: The intent resolution should be simplified for now:
The intent handling part of the Desktop Agent backend should define an async method for handling resolution. This method for now would select the first available running app, and if none of the running apps are available, launch the first appropriate app.
Later this method would be responsible for the resolver ui process.

ZKRobi (dismissed):

kruplm (dismissed):

BalassaMarton (approved):

kruplm (approved):

PR #418, opened 2024-01-26
Native support

ZKRobi:


src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader/Runners/NativeModuleRunner.cs
```diff
@@ -10,19 +10,85 @@
...
+        // Detach unexpected stop callbacks
+        foreach (var handler in moduleInstance.GetProperties<UnexpectedStopCallback>())
+        {
+            mainProcess.Exited -= handler.ProcessStoppedUnexpectedly;
+        }
+        var killNecessary = true;
+
+        if (mainProcess.CloseMainWindow())
+        {
+            await Task.Delay(TimeSpan.FromMilliseconds(500));
```
BalassaMarton, line 81: This delay seems arbitrary (and too short), what if the application pops up a dialog before quitting, or just takes a longer time to shut down?

src/messaging/dotnet/src/Client/Client/WebSocket/MessageRouterBuilderWebSocketExtensions.cs
```diff
@@ -1,37 +0,0 @@
-﻿// Morgan Stanley makes this available to you under the Apache License,
```
BalassaMarton, line 1: Why did you delete this file (and move its WS-specific methods to the more general builder class)?

src/shell/dotnet/Shell/MainWindow.xaml.cs
```diff
@@ -107,6 +106,15 @@ public ModuleViewModel(IModuleManifest manifest, ImageSourceProvider imageSource
                         webManifestDetails.Url);
                 }
             }
+            else if (manifest.TryGetDetails<NativeManifestDetails>(out var nativeManifestDetails))
+            {
+                var icon = System.Drawing.Icon.ExtractAssociatedIcon(Path.GetFullPath(nativeManifestDetails.Path.ToString()));
```
BalassaMarton, line 111: `// TODO: Let the manifest specify the icon`

src/messaging/dotnet/src/Client/DependencyInjection/MessageRouterBuilder.cs
```diff
@@ -11,17 +11,45 @@
...
+    public MessageRouterBuilder UseWebSocket(MessageRouterWebSocketOptions options)
+    {
+        ServiceCollection.AddSingleton<IOptions<MessageRouterWebSocketOptions>>(options);
+        ServiceCollection.AddSingleton<IConnectionFactory, WebSocketConnectionFactory>();
         return this;
     }
 
+    public MessageRouterBuilder UseWebSocketFromEnvironment()
+    {
+        var messageRouterUri = Environment.GetEnvironmentVariable("COMPOSEUI_MESSAGEROUTER_SERVER_URI");
```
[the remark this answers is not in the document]
ZKRobi: I knew I forgot something I wanted to do after a quick and dirty solution that doesn't worry about dependencies :)

src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader/Runners/NativeModuleRunner.cs
ZKRobi: True, it is completely arbitrary, and there are also thoughts that are not scope here to support more sophisticated shutdown flows (e.g. prompts to save) - I'll gladly change it to something more appropriate, but we should discuss what should be changed, I didn't want to change the interface without discussing.

src/messaging/dotnet/src/Client/Client/WebSocket/MessageRouterBuilderWebSocketExtensions.cs
ZKRobi: It seemed to me as if the extensions were remnants of an earlier version as they are in the same project it seemed like extra complexity to separate them like this.
I can revert this change, I don't have strong feelings, I didn't encounter this code organizing style before.

BalassaMarton: This pattern is everywhere in the .NET ecosystem, it improves composability and makes it easier to extract specific components into separate packages. Indeed, there was no indication that this was the case here, but please revert as the whole websocket protocol binding might (should?) get moved to its own package.

src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader.Abstractions/IModuleInstance.cs
```diff
@@ -37,4 +37,11 @@ public interface IModuleInstance
     /// </summary>
     /// <returns>The collection of properties in the order they were added to the <see cref="StartupContext"/></returns>
     IEnumerable<object> GetProperties();
+
+    /// <summary>
+    /// Gets the properties matching the specified type to the module instance
+    /// </summary>
+    /// <typeparam name="T">The type of the requested properties</typeparam>
+    /// <returns>An enumerable of the properties matching or castable to the desired type</returns>
+    IEnumerable<T> GetProperties<T>();
```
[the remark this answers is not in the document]
ZKRobi: Good point, moved to an extension class

src/messaging/dotnet/src/Client/Client/WebSocket/MessageRouterBuilderWebSocketExtensions.cs
ZKRobi: Reverted/moved the websocket-specific overload to the extension class

src/shell/dotnet/Shell/MainWindow.xaml.cs
ZKRobi: I added a property to the nativemanifestdetails, but it wouldn't work with the current implementation of imagesourceprovider, so I'm not adding it here yet.

src/messaging/dotnet/src/Client/Client/WebSocket/WebSocketEnvironmentVariables.cs
```diff
@@ -0,0 +1,18 @@
...
+// to in writing, software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions
+// and limitations under the License.
+
+namespace MorganStanley.ComposeUI.Messaging.Client.WebSocket;
+
+public static class WebSocketEnvironmentVariables
+{
+    public const string UriEnvironmentVariableName = "COMPOSEUI_MESSAGEROUTER_WEBSOCKET_URI";
```
ztanczos, line 17: Nitpick: since the class is called `WebSocketEnvironmentVariables` the `EnvironmentVariable` is redundant in `UriEnvironmentVariableName`.

ztanczos (approved):

BalassaMarton (dismissed):

BalassaMarton (approved):

PR #420, opened 2024-02-01
Fix existing standalone web examples

ztanczos:
- Add js-grid example to module catalog
- Remove redundant AppNotFoundException since that's available from MorganStanley.Fdc3.AppDirectory
- When the shell is started with the --url argument then create a ModuleManifest on-the-fly which is added to the ModuleCatalog so that it can be started via the ModuleLoader
- Serialize the WebWindowOptions to JSON and add it to the StartupRequest


BalassaMarton (dismissed):

kruplm (approved):

src/shell/dotnet/Shell/Modules/ModuleCatalog.cs
```diff
@@ -133,4 +116,21 @@ private struct ManifestTypeHelper
             public string ModuleType { get; set; }
         }
     }
+
+    internal class ModuleManifest : IModuleManifest
```
lilla28, line 120: Shouldn't we move this to a new file?

src/shell/dotnet/Shell/WebWindowOptions.cs
```diff
@@ -37,4 +37,5 @@ public sealed class WebWindowOptions
     public const string DefaultTitle = "Compose Web Container";
     public const string DefaultUrl = "about:blank";
     public const double DefaultWidth = 800;
+    public const string ParameterName = nameof(WebWindowOptions);
```
lilla28, line 40: This name is a little bit weird for me, but I'll accept it.

src/shell/dotnet/Shell/Modules/WebWindowOptionsStartupAction.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public Task InvokeAsync(StartupContext startupContext, Func<Task> next)
+        {
+            if (startupContext.ModuleInstance.Manifest.ModuleType == ModuleType.Web)
+            {
+                var webWindowOptions = startupContext.StartRequest.Parameters
+                    .FirstOrDefault(p => p.Key == WebWindowOptions.ParameterName).Value;
+                if (webWindowOptions != null)
+                {
+                    startupContext.AddProperty(JsonSerializer.Deserialize<WebWindowOptions>(webWindowOptions));
```
BalassaMarton, line 31: You'll have to update the logic here to use the injected `WebWindowOptions` if there is one:
https://github.com/morganstanley/ComposeUI/blob/main/src/shell/dotnet/Shell/Modules/ModuleService.cs#L74

BalassaMarton (changes requested):

src/shell/dotnet/Shell/Modules/WebWindowOptionsStartupAction.cs
ztanczos: That's what I do here: https://github.com/morganstanley/ComposeUI/pull/420/files#diff-80658622f21af86e1da2cc17cdf2c25b2d84074dd125a2dbb9ec3408883bac83R66

BalassaMarton: Oh OK for some reason I saw a previous version of that file

BalassaMarton (approved):

BalassaMarton (dismissed):

PR #463, opened 2024-02-09
fix(Examples/WPFDiagnosticsExample): Check if Executable Exists

kruplm:
**Changes in this PR**
* Adding Error Handling for extracting the icon from a native app

README is updated in https://github.com/morganstanley/ComposeUI/pull/462

### DIFF
--- a/src/shell/dotnet/Shell/MainWindow.xaml.cs
+++ b/src/shell/dotnet/Shell/MainWindow.xaml.cs
@@ -106,13 +106,25 @@ public partial class MainWindow : RibbonWindow
             }
             else if (manifest.TryGetDetails<NativeManifestDetails>(out var nativeManifestDetails))
             {
-                var icon = System.Drawing.Icon.ExtractAssociatedIcon(Path.GetFullPath(nativeManifestDetails.Path.ToString()));
-
-                ImageSource = System.Windows.Interop.Imaging.CreateBitmapSourceFromHIcon(
-                icon.Handle,
-                new Int32Rect { Width = icon.Width, Height = icon.Height },
-                BitmapSizeOptions.FromEmptyOptions());
+                var icon = GetNativeAppIcon(nativeManifestDetails);
+                if (icon != null) { 
+                    ImageSource = System.Windows.Interop.Imaging.CreateBitmapSourceFromHIcon(
+                    icon.Handle,
+                    new Int32Rect { Width = icon.Width, Height = icon.Height },
+                    BitmapSizeOptions.FromEmptyOptions());
+                }
             }
+        }
+
+        public System.Drawing.Icon? GetNativeAppIcon(NativeManifestDetails nativeManifestDetails) {
+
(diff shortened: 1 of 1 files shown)


src/shell/dotnet/Shell/MainWindow.xaml.cs
```diff
@@ -106,12 +106,24 @@ public ModuleViewModel(IModuleManifest manifest, ImageSourceProvider imageSource
             }
             else if (manifest.TryGetDetails<NativeManifestDetails>(out var nativeManifestDetails))
             {
-                var icon = System.Drawing.Icon.ExtractAssociatedIcon(Path.GetFullPath(nativeManifestDetails.Path.ToString()));
+                var icon = GetNativeAppIcon(nativeManifestDetails);
+                if (icon != null) { 
+                    ImageSource = System.Windows.Interop.Imaging.CreateBitmapSourceFromHIcon(
```
psmulovics, line 111: Can we check whether we need to release unmanaged resources like handles after this?

psmulovics (approved):

src/shell/dotnet/Shell/MainWindow.xaml.cs
ZKRobi: Good catch, an Icon is disposable

psmulovics: It is unrelated to this change - I would say go ahead and merge this after switching to using and create a new item for to check for losing handles. Want to avoid having this open too long.

src/shell/dotnet/Shell/MainWindow.xaml.cs
```diff
@@ -106,13 +106,25 @@ public ModuleViewModel(IModuleManifest manifest, ImageSourceProvider imageSource
...
+            }
+        }
 
-                ImageSource = System.Windows.Interop.Imaging.CreateBitmapSourceFromHIcon(
-                icon.Handle,
-                new Int32Rect { Width = icon.Width, Height = icon.Height },
-                BitmapSizeOptions.FromEmptyOptions());
+        public System.Drawing.Icon? GetNativeAppIcon(NativeManifestDetails nativeManifestDetails) {
+
+           var path = Path.GetFullPath(nativeManifestDetails.Path.ToString());
```
ZKRobi, line 121: As we discussed, if a relative path is provided for a (native) application, we will understand that path relative to the manifest file (instead of the working path as it is done here, and in the runner). We think this is easier to understand for the creator of a manifest, more flexible for qa or debugging purposes, and more deterministic in the average use-case.

An example would be the following:
The ComposeUI based app is running from the folder c:\ComposeUIApp
The manifest file is located under c:\Modules\manifest.json

The manifest contains an app with the path .\myApp\myApp.exe

The current implementation would try to find the icon and launch .\ComposeUIApp\myApp\myApp.exe
The expectation instead would be .\Modules\myApp\myApp.exe

PR #465, opened 2024-02-20
Initial PR for NX monorepo

lilla28:
**Changes:**
- Added nx/dotnet plugin to the monorepo so we can build/test/serve our applications/libraries
- Removed Lerna from the monorepo
- Added project.json files
- Fixed build scripts: NX<-Lerna
- Modified tests so they can start processes from the dist folder, as now this is the output folder even though if you are building from VS22

**Known Issues:**
1. Sometimes the tests are getting stuck
2. Sometimes the build is corrupted: (remove package.lock.json and node_modules from the root and run npm i again)
3. Sometimes tests are randomly failing for example: "EXEC : error : EPERM: operation not permitted, rename ...", and sometimes not :) but if you run them individually they will pass.
---->UPDATED: Some sources suggests to clean npm cache, because it might be corrupted (they expreienced this kind of errors during npm i)
 (not specific to NX, but https://bobbyhadz.com/blog/npm-err-code-eperm-operation-not-permitted)
 It also could be file permissions problem, anti-virus problem, temporary directory permissions problem, old dependencies(it is using the latest NX version)


kruplm (approved):

.github/workflows/release.yml
```diff
@@ -58,9 +58,8 @@ jobs:
       - uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
         with:
           node-version: 20.x
-          registry-url: https://registry.npmjs.org/
       - run: |
-             lerna publish from-package --no-private --yes
+             npx nx release publish --dry-run --projects=tag:release # THIS IS DRY-RUN
```
kruplm, line 62: looks very promising, thanks for addressing this

PR #471, opened 2024-02-27
chore(messaging): Added Publish-, Subscribe- and UnsubscribeResponse types on both .NET and Typescript side, fixed DesktopAgent ModuleNotFound exception when removing module from the dictionary -running the chart and grid example.

lilla28:
- Added Publish-, Subscribe- and UnsubscribeResponse types on both .NET and Typescript side
- Fixed DesktopAgent ModuleNotFound exception when removing module from the dictionary -running the chart and grid example.


src/messaging/dotnet/src/Server/Server/MessageRouterServer.cs
```diff
@@ -322,11 +356,36 @@ public ValueTask ClientDisconnected(IClientConnection connection)
...
+                },
+                CancellationToken.None);
+
+            OnRequestStop(message);
         }
-        finally
+        catch (Exception exception)
         {
-            OnRequestStop(message);
+            if (_logger.IsEnabled(LogLevel.Debug))
```
ztanczos, line 370: I'd log exceptions on Error level

src/messaging/dotnet/src/Server/Server/MessageRouterServer.cs
```diff
@@ -322,11 +356,36 @@ public ValueTask ClientDisconnected(IClientConnection connection)
...
+            try
+            {
+                await client.Connection.SendAsync(
+                    new SubscribeResponse
+                    {
+                        RequestId = message.RequestId,
+                        Error = new Error(exception)
+                    },
+                    CancellationToken.None);
+            }
```
ztanczos, line 384: Shouldn't we `throw` before returning?

src/messaging/js/composeui-messaging-client/src/client/MessageRouterClient.ts
```diff
@@ -83,11 +83,16 @@ export class MessageRouterClient implements MessageRouter {
...
-                    topic: topicName
-                });
+            try {
+                await this.sendRequest<messages.SubscribeMessage, messages.SubscribeResponse>(
+                    {
+                        requestId: this.getRequestId(),
+                        type: "Subscribe",
+                        topic: topicName
+                    });
+            } catch (error) {
```
ztanczos, line 93: What's the point of just throwing in a catch block?

src/messaging/js/composeui-messaging-client/src/client/MessageRouterClient.ts
```diff
@@ -96,13 +101,18 @@ export class MessageRouterClient implements MessageRouter {
...
+            await this.sendRequest<messages.PublishMessage, messages.PublishResponse>(
+                {
+                    type: "Publish",
+                    requestId: this.getRequestId(),
+                    topic,
+                    payload,
+                    correlationId: options?.correlationId
+                });
+        } catch (error) {
+            throw error;
```
bingenito, line 114: This is rewriting the exception stack to root here, is that intentional?

src/messaging/js/composeui-messaging-client/src/client/MessageRouterClient.ts
bingenito: It's also rewriting the exception stack to root here, is that intentional?

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -695,10 +695,17 @@ Task<bool> IsIntentListenerRegisteredAsync(AppMetadata appMetadata)
...
+            var fdc3InstanceId = GetFdc3InstanceId(instance);
+            if (!_runningModules.TryRemove(new(fdc3InstanceId), out _))
+            {
+                _logger.LogError($"Could not remove the closed window with instanceId: {fdc3InstanceId}.");
+            }
+        }
+        catch (Fdc3DesktopAgentException exception)
         {
-            _logger.LogError($"Could not remove the closed window with instanceId: {fdc3InstanceId}.");
+            _logger.LogError(exception, $"Exception thrown while removing module: {instance.Manifest.Id}, {instance.Manifest.Name} from running instances in FDC3DesktopAgent.");
```
bingenito, line 708: Should we be making of use ? in here to guard against nulls in a catch logging statement?

src/messaging/dotnet/src/Client/Client/MessageRouterClient.cs
```diff
@@ -797,11 +799,21 @@ async Task CloseTopics()
...
+        try
+        {
+            if (topic.CanUnsubscribe)
+            {
+                await SendRequestAsync(new UnsubscribeMessage { RequestId = requestId, Topic = topic.Name }, CancellationToken.None);
+            }
+        }
+        catch (MessageRouterException exception)
+        {
+            _logger.LogWarning(exception, $"Exception thrown while unsubscribing, topic: {topic.Name}, request id: {requestId}.");
```
bingenito, line 815: Should we be making of use ? in here to guard against nulls in a catch logging statement?

src/messaging/dotnet/src/Server/Server/MessageRouterServer.cs
```diff
@@ -322,11 +356,36 @@ public ValueTask ClientDisconnected(IClientConnection connection)
...
+
+            OnRequestStop(message);
         }
-        finally
+        catch (Exception exception)
         {
-            OnRequestStop(message);
+            if (_logger.IsEnabled(LogLevel.Debug))
+            {
+                _logger.LogDebug($"Exception thrown while handling subscription message: {0}.", exception);
```
bingenito, line 372: Are there not overloads that take an exception as a parameter?  Do we really need to create strings on the heap?

src/messaging/dotnet/src/Server/Server/MessageRouterServer.cs
```diff
@@ -307,7 +341,7 @@ public ValueTask ClientDisconnected(IClientConnection connection)
         try
         {
             if (!Protocol.Topic.IsValidTopicName(message.Topic))
```
BalassaMarton, line 343: Replace with `Protocol.Topic.Validate(message.Topic)`. It will throw an exception that you can send back to the client in the `catch` arm.

src/messaging/dotnet/src/Server/Server/MessageRouterServer.cs
```diff
@@ -322,11 +356,36 @@ public ValueTask ClientDisconnected(IClientConnection connection)
...
+                    new SubscribeResponse
+                    {
+                        RequestId = message.RequestId,
+                        Error = new Error(exception)
+                    },
+                    CancellationToken.None);
+            }
+            finally
+            {
+                OnRequestStop(message);
```
BalassaMarton, line 387: `OnRequestStop(message, exception)`

src/messaging/dotnet/src/Server/Server/MessageRouterServer.cs
BalassaMarton: No. That would crash the whole connection. Just logging the error is fine.

src/messaging/dotnet/src/Server/Server/MessageRouterServer.cs
```diff
@@ -378,12 +437,35 @@ public ValueTask ClientDisconnected(IClientConnection connection)
...
+                await client.Connection.SendAsync(
+                    new UnsubscribeResponse
+                    {
+                        RequestId= message.RequestId,
+                        Error = new Error(exception)
+                    }, CancellationToken.None);
+            }
+            finally
+            {
+                OnRequestStop(message);
```
BalassaMarton, line 467: `OnRequestStop(message, exception)`

src/messaging/js/composeui-messaging-client/src/client/MessageRouterClient.ts
BalassaMarton: That's not a bad idea, the error object is created by `sendRequestAsync`, but it makes more sense to see the current method in the stack trace, as if the error was thrown here.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -695,10 +695,17 @@ Task<bool> IsIntentListenerRegisteredAsync(AppMetadata appMetadata)
...
-        if (!_runningModules.TryRemove(new(fdc3InstanceId), out _))
+        try
+        {
+            var fdc3InstanceId = GetFdc3InstanceId(instance);
+            if (!_runningModules.TryRemove(new(fdc3InstanceId), out _))
+            {
+                _logger.LogError($"Could not remove the closed window with instanceId: {fdc3InstanceId}.");
+            }
+        }
+        catch (Fdc3DesktopAgentException exception)
```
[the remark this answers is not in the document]
lilla28: This exception is catched because if that is not a fdc3 enabled app aka it has no fdc3instanceId then we shouldn't care if the GetFdc3InstanceId(IModuleInstance) throws an error as the fdc3InstanceId is missing. If you start the messagerouter examples: chart and grid then you experiment this issue with the current shell on close if you set the exception settings to catch all.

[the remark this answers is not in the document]
ztanczos: If it's not an FDC3 app then why are we even here in Fdc3DesktopAgent?

src/messaging/dotnet/src/Client/Client/MessageRouterClient.cs
lilla28: topic, requestId and exception are not null here, I do not see the point to guard here. but if you insist I can add. :)

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
lilla28: Same here, the variables/parameters shouldn't be null values.

src/messaging/dotnet/src/Client/Client/MessageRouterClient.cs
bingenito: topic is never checked for null.  The first  access to it within the try can throw and then the log here will also fail if null

src/messaging/js/composeui-messaging-client/src/client/MessageRouterClient.ts
lilla28: My original idea was that. I will remove the try catch block then from the related calls I modified.

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
lilla28: `RemoveModuleAsync` is also called from a subscribe event, when a module is stopped fired by the moduleLoader itself, so in this case the `IModuleInstance` and it's manifest's id and name shouldn't be null

src/fdc3/dotnet/DesktopAgent/src/DesktopAgent/Fdc3DesktopAgent.cs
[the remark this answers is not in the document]
ztanczos: So once we enable FDC3 we can no longer run non-FDC3 applications? That sounds bad, and the solution won't be an empty `catch` block.
For now I'm OK with this change because it is not related to your main change, but there should be a follow-up to fix this properly.

ztanczos (dismissed):

kruplm:
Also when you fix the imports above, could you please include this as well?
https://github.com/lilla28/ComposeUI/blob/messaging/response-types/src/messaging/js/composeui-messaging-client/src/protocol/messages/RegisterServiceRequest.ts#L14 

**import { EndpointDescriptor } from "../../EndpointDescriptor";**

I know it's not new code, but I spot this during the review as well.
Thank you

kruplm (approved):

ztanczos (approved):

ZKRobi (approved):

PR #540, opened 2024-03-08
chore(actions): update default token permissions to read

bingenito:


.github/workflows/release.yml
```diff
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  contents: write
+  contents: read
```
bingenito, line 10: Documentation for upload of release artifact has no mention of requiring write

psmulovics (approved):

PR #557, opened 2024-03-19
Feature/nuget publish test

ZKRobi:
This PR is only targeting a feature branch to test a "dry run" of the nuget publish and no other release steps. As there is no dry run option on pushing to nuget, I only do a "dir" command on the folder the desired packages are expected to be to verify everything is ready for the actual push.

This will never be merged to the main branch, after the test the feature branch should be deleted. A separate PR will be raised with the real changes based on the results of this test.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -49,47 +49,17 @@ jobs:
 
     - name: Install NuGet dependencies
       run: |
-        $failedSolutions = @()
-        
-        Get-ChildItem -Recurse -Include *.sln `
-        | ForEach-Object {
-            dotnet restore $_;
-            if ($LASTEXITCODE -ne 0) {$failedSolutions += Split-Path $_ -leaf; }
-          }
-        
-        if ($failedSolutions.count -gt 0) {
-            throw "Restore FAILED for solutions $failedSolutions"
-        }
+           ./build/dotnet-restore.ps1
 
 
     - name: Build .NET
       run: |
-        $failedSolutions = @()
-        
-        Get-ChildItem -Recurse -Include *.sln `
-        | ForEach-Object {
-            dotnet build $_ --configuration Release --no-restore; 
-            if ($LASTEXITCODE -ne 0) {$failedSolutions += Split-Path $_ -leaf; }
-          }
-        
-        if ($failedSolutions.count -gt 0) {
-            throw "Build FAILED for solutions $failedSolutions"
-        }
+           powershell ./build/dotnet-build.ps1
 
 
     - name: Test .NET
       run: |
-        $failedSolutions = @()
-        
-        Get-ChildItem -Recurse -Include *.sln `
-        | ForEach-Object {
-            dotnet test $_ --configuration Release --no-build; 
-            if ($LASTEXITCODE -ne 0) { $failedSolutions += Split-Path $_ -leaf; }
-        }
-        
-        if ($failedSolutions.count -gt 0) {
-            throw "Test FAILED for solutions $failedSolutions"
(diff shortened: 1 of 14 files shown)


.github/workflows/release.yml
```diff
@@ -15,54 +15,21 @@ name: Release
...
+      - uses: actions/download-artifact@c850b930e6ba138125429b7e5c93fc707a7f8427 # v4.1.4
         with:
-          node-version: 18
-          registry-url: https://registry.npmjs.org/
-      - run: |
+        name: packages
+        path: ./packages
+      - uses: actions/setup-dotnet@4d6c8fcf3c8f7a60068d26b594648e99df24cee3 # v4.0.0
+        with:
+          dotnet-version: ${{ matrix.dotnet-version }}
```
bingenito, line 30: I don't see a matrix defined in here but also, why a matrix, wouldn't we be overwriting each package?

.github/workflows/release.yml
```diff
@@ -15,54 +15,21 @@ name: Release
 jobs:
   build:
     uses: ./.github/workflows/continuous-integration.yml
-  upload:
```
bingenito, line 18: What happened to all of this?  Aren't the npm packages for shell binaries also being uploaded?

ZKRobi: This PR is targeting a feature branch only to test that the build/pack output is correctly in place. This version will not ever be merged to the main branch. I removed all the npm stuff to prevent uploading npm packages and the zip file during the test run.

.github/workflows/release.yml
ZKRobi: Ideally I'd like to have a central place to define what versions are we using instead of defining .net (and node) versions all over the place in order to reduce the cognitive load and the chance for human error. Using a matrix seemed to be the closest I can get to this without ugly hacks, and I saw Maria using it before. I think I got lost in the scopes of things though so let me revisit this idea :)

ZKRobi: Sidenote: I think I found a way to achieve this without hacking, but that's out of scope for this, so I just hardcoded the .net version for now

.github/workflows/release.yml
```diff
@@ -15,54 +15,21 @@ name: Release
...
         with:
-          node-version: 18
-          registry-url: https://registry.npmjs.org/
-      - run: |
+        name: packages
+        path: ./packages
+      - uses: actions/setup-dotnet@4d6c8fcf3c8f7a60068d26b594648e99df24cee3 # v4.0.0
+        with:
+          dotnet-version: 6.0.x
+      - name: Publish Messaging
```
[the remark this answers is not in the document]
ZKRobi: Sure I'll add it to the real version :)

.github/workflows/continuous-integration.yml
```diff
@@ -49,57 +49,35 @@ jobs:
...
-        Get-ChildItem -Recurse -Include *.sln `
-        | ForEach-Object {
-            dotnet restore $_;
-            if ($LASTEXITCODE -ne 0) {$failedSolutions += Split-Path $_ -leaf; }
-          }
-        
-        if ($failedSolutions.count -gt 0) {
-            throw "Restore FAILED for solutions $failedSolutions"
-        }
+           ./build/dotnet-restore.ps1
```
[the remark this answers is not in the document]
ZKRobi: fixed

kruplm (approved):

.github/workflows/release.yml
bingenito: I think you would move the setup-dotnet job into its own yml and have it as a required of the other actions

PR #561, opened 2024-03-20
Publishing Messaging nuget packages and some related CI streamlining #560

ZKRobi:
In order to publish the nuget packages I decided some streamlining on the dotnet build side is necessary. I separated the changes in 3 commits so they are easier to review and follow in the history.

I made the existing CI job to make use of the build scripts in the repository. This change is intended to have the ability to run the same build process locally.
I added variable usage for DOTNET_VERSION and NODE_VERSION so we don't have to adjust this across multiple workflow files. This needs to be set up on the main repo as well.
I adjusted the packaging scripts and projects so by default projects are not packaged, version is handled in a central location and packages are output in a central location. I enabled packaging for the Messaging libraries where necessary.
I added a deploy_to_nuget job to the release.yml workflow that pushes the packages. Note that packages are intentionally listed one by one in order to avoid accidentally pushing undersired packages to nuget.org
Further streamlining of the processes might be desirable, I didn't want to inflate this PR further.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,7 @@ jobs:
         run: Compress-Archive -Path  ${{ github.workspace }}/shell-binaries/* composeui-${{ github.ref_name }}-win32.zip
         
       - name: Upload Release Asset
-        id: upload-release-asset 
+        id: upload-release-asset
         uses: actions/upload-release-asset@e8f9f06c4b078e705bd2ea027f0926603fc9b4d5 # v1.0.2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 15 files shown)


.github/workflows/release.yml
```diff
@@ -64,5 +64,29 @@ jobs:
              npm whoami
              npm publish --access public
         env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         working-directory: ${{ github.workspace }}/src/shell/js/composeui-node-launcher/
+
+  deploy_to_nuget:
```
lilla28, line 70: I know it's picky but it's a little bit inconsistent naming:
deploy-npm vs deploy_to_nuget

ZKRobi: Ah yeah in the original version I made I named both to matchin patterns but I didn't change in this final version, I'll adjust :)

.github/workflows/release.yml
```diff
@@ -53,12 +53,37 @@ jobs:
...
         with:
-          node-version: 18
+          node-version: ${{ vars.NODE_VERSION }}
           registry-url: https://registry.npmjs.org/
       - run: |
              lerna publish from-package --no-private --yes
         env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        working-directory: ${{ github.workspace }}/src/shell/js/composeui-node-launcher/
```
kruplm, line 66: I've removed the working directory from here.
https://github.com/morganstanley/ComposeUI/pull/563/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34

If seems that it's left here after the merge conflict resolution.
Please remove it, lerna publish has to run on the whole monorepo.

kruplm (changes requested):

kruplm (approved):

kruplm (changes requested):
It doesn't seem to pick up the version upgrade for Node.js 20 from the repo variable despite updating the value:
Environment details
  node: v18.18.2
  npm: 9.8.1
  yarn: 1.22.22
  
I recommend putting back the matrix for now as provenance won't happen for the NPM packages this way.

kruplm (approved):

PR #569, opened 2024-03-25
fix(module-loader) - Fixed ModuleLoader behavior with multiple module catalogs

lilla28:
Changes:
- picked up changes from branch and commits related to the ModuleLoader when it tries to register its ModuleCatalog https://github.com/morganstanley/ComposeUI/commit/89ec2d9e1f9be47df0cacfdbb4586d9405d0b332 && https://github.com/morganstanley/ComposeUI/commit/57ad7daab2f8e28d63b335562b7f8b2bdf20fe79
- fixed tests when it registers its modulecatalogs (mocking array of catalogs)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7247,8 +7247,12 @@
       "dev": true
     },
     "node_modules/@types/node": {
-      "version": "18.14.6",
-      "license": "MIT"
+      "version": "20.12.6",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-20.12.6.tgz",
+      "integrity": "sha512-3KurE8taB8GCvZBPngVbp0lk5CKi8M9f9k1rsADh0Evdz5SzJ+Q+Hx9uHoFGsLnLnd1xmkDQr2hVhlA0Mn0lKQ==",
+      "dependencies": {
+        "undici-types": "~5.26.4"
+      }
     },
     "node_modules/@types/normalize-package-data": {
       "version": "2.4.4",
@@ -19018,6 +19022,11 @@
         "node": ">=18.0"
       }
     },
+    "node_modules/undici-types": {
+      "version": "5.26.5",
+      "resolved": "https://registry.npmjs.org/undici-types/-/undici-types-5.26.5.tgz",
+      "integrity": "sha512-JlCMO+ehdEIKqlFxk6IfVoAUVmgz7cU7zD/h9XZ0qzeosSHmUJVOzSQvvYSYWXkFXC+IfLKSIffhv0sVZup6pA=="
+    },
     "node_modules/unicode-canonical-property-names-ecmascript": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/unicode-canonical-property-names-ecmascript/-/unicode-canonical-property-names-ecmascript-2.0.0.tgz",
@@ -20495,7 +20504,7 @@
       "dependencies": {
         "@finos/fdc3": "^2.0.0",
         "@morgan-stanley/composeui-messaging-client": "*",
-        "@types/node": "^18.11.18",
+        "@types/node": "^20.12.6",
         "jest-environment-jsdom": "^29.7.0",
         "rxjs": "^7.8.1"
       },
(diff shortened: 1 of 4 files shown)


kruplm (dismissed):

ztanczos (approved):

fhubi (approved):

src/module-loader/dotnet/src/MorganStanley.ComposeUI.ModuleLoader/ModuleLoader.cs
```diff
@@ -106,4 +102,50 @@ private async Task StopModuleInternal(IModuleInstance moduleInstance)
 
         _lifetimeEvents.OnNext(new LifetimeEvent.Stopped(moduleInstance));
     }
+
+    private class AggregateModuleCatalog : IModuleCatalog
```
ZKRobi, line 106: Please consider moving this class to it's own file as it implements complex behavior. I'd also consider unit testing it, although integration testing may cover all use-cases. That said, I don't see the multi-catalog scenario covered by tests, and I don't see what happens if multiple catalogs provide an entry with the same moduleId (perhaps with different contents).

ZKRobi (changes requested):

src/module-loader/dotnet/tests/MorganStanley.ComposeUI.ModuleLoader.Tests/AggregateModuleCatalogTests.cs
```diff
@@ -0,0 +1,147 @@
...
+    [InlineData("testModuleId8")]
+    public async Task GetManifest_ThrowsModuleNotFoundException_WhenNoModuleIdFound(string moduleId)
+    {
+        var action = () => _moduleCatalog.GetManifest(moduleId);
+        await action.Should().ThrowAsync<ModuleNotFoundException>();
+    }
+
+    [Theory]
+    [InlineData("testModuleId")]
+    public async Task GetManifest_ReturnsTheFirstOccurrence_WhenMultipleCatalogContainsTheSameModuleId(string moduleId)
```
ZKRobi, line 67: The same Id exists with the same name multiple times. This doesn't verify the original source for the manifest as it can return any of the occurances and the assertion will stay true.

ZKRobi (changes requested):

architecture/adr-014-module-catalog.md
```diff
@@ -0,0 +1,21 @@
+<!-- Morgan Stanley makes this available to you under the Apache License, Version 2.0 (the "License"). You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0. See the NOTICE file distributed with this work for additional information regarding copyright ownership. Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License. -->
+
+| Type          | ID            | Status        | Title                  |
+| ------------- | ------------- | ------------- | ---------------------- |
+| ADR           | adr-014       | Proposed      | Multiple ModuleCatalog |
```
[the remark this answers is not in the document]
lilla28: I will create a new commit with the change of the status once everything is resolved and we discussed.

architecture/adr-014-module-catalog.md
```diff
@@ -0,0 +1,21 @@
...
+# Architecture Decision Record: Multiple ModuleCatalog
+
+## Context
+ComposeUI aims to support many different kind of modules and applications. These applications/modules, that can be started via the ModuleLoader should be stored in ModuleCatalogs and every module should have it's unique module id. The user should have the ability to handle multiple ModuleCatalog with the ModuleLoader.
+
+General problem statement:
+
+- When multiple ModuleCatalog is being used, some modules could have the same module id in different ModuleCatalog which could cause issue while starting/stopping the application/module via the ModuleLoader.
+
+## Decision
```
ztanczos, line 17: Just an idea: what if we'd fail fast & hard if there are duplicates, i.e.: throw an exception and crash? What do you think?

ZKRobi: To recap, the issue with failing on duplicates from different sources is that it might cause a situation that the developer/user has no way of resolving.
Ideally this is never an issue but if merging multiple module catalogs is a feature we want to support and we wish to keep the standard interface (and not do something like e.g. prefixing ids that would be non-standard), we need a well documented consistent way of resolving these conflicts that the developer/user has a way to control.

ztanczos (dismissed):

kruplm (approved):

PR #610, opened 2024-04-16
fix(icon-handling) - Disposing icon after ImageSource is created from its handle

lilla28:
Changes:
- Added using statement when we use the Icon's Handle to create ImageSource. So after the ImageSource is created the icon is disposed.


src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -121,11 +120,11 @@ private async Task StartAsync(StartupEventArgs e)
         var diagnostics = new DiagnosticInfo
         {
             StartupTime = DateTime.Now,
-            ShellVersion = Assembly.GetExecutingAssembly().FullName
+            ShellVersion = Assembly.GetExecutingAssembly().FullName ?? throw new ArgumentNullException(nameof(Assembly.GetExecutingAssembly))
```
ztanczos, line 123: `ArgumentNullException` should be thrown when a required method argument is null.
What do you want to achieve here?

PR #640, opened 2024-05-07
fix(messaging-documentation) - Fix Javascript Gatsby highlights

lilla28:
Changes:
- Fixed gatsby highlights for Javascript codes as well


kruplm (approved):

kruplm:
The markup is not case-sensitive, but let's merge it for consistency with "csharp" being lower-case :)

PR #744, opened 2024-08-22
FDC3 improvements

lilla28:
Implemented:
getOrCreateChannel
getUserChannels
getInfo
findInstances
getAppMetadata

Added some implementation to track the joined userchannels properly via fdc3.joinUserChannels -> messagerouter.invoke calls, todos added for the restriction of joining based on the userchannel set.

Refactored raiseIntent, findIntent, findIntentsByContext based on the FDC3 Conformance tests -> added exception handling based on the tests, quering based on the intentmetadata's key instead of the Name property


ztanczos (dismissed):

src/fdc3/dotnet/DesktopAgent/test/MorganStanley.ComposeUI.DesktopAgent.Tests/Channels/ChannelTestFixture.cs
```diff
@@ -15,94 +15,92 @@
 using Finos.Fdc3.Context;
 using MorganStanley.ComposeUI.Fdc3.DesktopAgent.Channels;
 using MorganStanley.ComposeUI.Fdc3.DesktopAgent.Contracts;
-using MorganStanley.ComposeUI.Messaging.Abstractions;
 
-namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Tests;
+namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Tests.Channels;
 
-public class UserChannelTests
+public abstract class ChannelTestFixture
```
ZKRobi, line 21: Wouldn't calling this ChannelTestBase be more appropriate?

src/fdc3/dotnet/DesktopAgent/test/MorganStanley.ComposeUI.DesktopAgent.Tests/Channels/UserChannelTests.cs
```diff
@@ -0,0 +1,31 @@
...
+
+public class UserChannelTests : ChannelTestFixture
+{
+    private const string TestChannel = "testChannel";
+    private readonly UserChannel _channel = new UserChannel(TestChannel, new Mock<IMessagingService>().Object, null);
+    private readonly  ChannelTopics _topics = Fdc3Topic.UserChannel(TestChannel);
+
+    public UserChannelTests()
+    {
+        Channel = _channel;
```
ZKRobi, line 28: This seems a bit odd, why not just create the UserChannel and the ChannelTopics here instead of creating a field that's never used?

src/fdc3/dotnet/DesktopAgent/test/MorganStanley.ComposeUI.DesktopAgent.Tests/EndToEndTests.cs
```diff
@@ -257,6 +257,8 @@ public async Task FindIntentReturnsAppIntent()
         var result = resultBuffer!.ReadJson<FindIntentResponse>(_options);
         result.Should().NotBeNull();
         result.Should().BeEquivalentTo(expectedResponse);
+
+        await _moduleLoader.StopModule(new StopRequest(instance.InstanceId));
```
ZKRobi, line 266: These will not be called if the assertion fails, it should either be in a try-finally block or better yet make the test class disposable and do the cleanup there - that would probably spare a lot of boilerplate code as well

src/fdc3/dotnet/DesktopAgent/src/MorganStanley.ComposeUI.DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -462,6 +485,68 @@ public async ValueTask<RaiseIntentResult<IntentListenerResponse>> AddIntentListe
...
+    }
+
+    public async ValueTask<JoinUserChannelResponse?> JoinUserChannel(UserChannel channel, string instanceId)
+    {
+        if (!Guid.TryParse(instanceId, out var id) || !_runningModules.TryGetValue(id, out _))
+        {
+            return JoinUserChannelResponse.Failed(ChannelError.AccessDenied);
+        }
+
+        //TODO remove if regarding this
```
ZKRobi: Any reason for not doing this?

src/fdc3/dotnet/DesktopAgent/src/MorganStanley.ComposeUI.DesktopAgent/Fdc3DesktopAgent.cs
```diff
@@ -462,6 +485,68 @@ public async ValueTask<RaiseIntentResult<IntentListenerResponse>> AddIntentListe
...
+        ChannelItem? channelItem = null;
+        if (_userChannelSet != null && !_userChannelSet.TryGetValue(channel.Id, out channelItem))
+        {
+            //TODO delete this if statement
```
ZKRobi: Any reason for not doing this?

src/fdc3/dotnet/DesktopAgent/src/MorganStanley.ComposeUI.DesktopAgent/Infrastructure/Internal/Constants.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System.Reflection;
+
+namespace MorganStanley.ComposeUI.Fdc3.DesktopAgent.Infrastructure.Internal;
+
+internal static class Constants
+{
+    public static string DesktopAgentProvider = "ComposeUI";
+    public static string SupportedFdc3Version = "2.0.0";
+    public static string? ComposeUIVersion = Assembly.GetExecutingAssembly().GetName().Version?.ToString();
```
ZKRobi, line 23: No point in using GetExecutingAssembly here, better to use typeof(Constants).Assembly (or any other type from the assembly)
See: https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getexecutingassembly?view=net-8.0#remarks

Not that this is a performance-wise critical area, but it's a good/bad practice thing.

src/fdc3/js/composeui-fdc3/src/infrastructure/ComposeUIIntentListener.ts
```diff
@@ -57,14 +59,15 @@ export class ComposeUIIntentListener implements Listener {
...
                         request = new Fdc3StoreIntentResultRequest(message.messageId, this.intent, this.instanceId, message.contextMetadata.source.instanceId!, undefined, undefined, undefined, true);
                     }
-
                 } catch (error) {
-                    request = new Fdc3StoreIntentResultRequest(message.messageId, this.intent, this.instanceId, message.contextMetadata.source.instanceId!, undefined, undefined, undefined, false, ResultError.IntentHandlerRejected);
+                    console.error(error);
+                    request = new Fdc3StoreIntentResultRequest(message.messageId, this.intent, this.instanceId, message.contextMetadata.source.instanceId!, undefined, undefined, undefined, false, error as string);
                 }
 
+                console.log("STORE INTENT RESULT REQUEST:", request);
```
ZKRobi: I'm not convinced leaving debug messages like this is good practice, but I won't argue that until we look at an improved logging. Please write it in without all caps though :)

ZKRobi (commented):
Please squash the follow-up commits into the appropriate original commits

src/fdc3/dotnet/DesktopAgent/src/MorganStanley.ComposeUI.DesktopAgent/Fdc3DesktopAgent.cs
lilla28: next pr :)

src/fdc3/dotnet/DesktopAgent/src/MorganStanley.ComposeUI.DesktopAgent/Fdc3DesktopAgent.cs
lilla28: next pr

examples/fdc3-trade-simulator/js-order-book/package-lock.json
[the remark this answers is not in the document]
lilla28: sorry that was automatically created when I have created the example app, I will remove it

src/fdc3/dotnet/DesktopAgent/test/MorganStanley.ComposeUI.DesktopAgent.Tests/EndToEndTests.cs
lilla28: We already using IAsyncLifetime which exposes that DisposeAsync will be called after every tests. I remove the unnecessary/redundant and wrong calls :)

src/fdc3/js/composeui-fdc3/src/infrastructure/ComposeUIIntentListener.ts
lilla28: My initial idea leaving those comments was to keep these so you can try out the bug we found - ContextListener is not receiving the messages (from console logs it turns out the message is being addressed etc), but you are right I'll remove these

ztanczos (dismissed):

ZKRobi (approved):

PR #806, opened 2024-10-15
Webpack build js chart grid example

kruplm:
Changed bundler from rollup to webpack in js-chart-and-grid-messagerrouter example


examples/js-chart-and-grid-messagerouter/js-chart/webpack.config.cjs
```diff
@@ -0,0 +1,40 @@
+// Generated using webpack-cli https://github.com/webpack/webpack-cli
```
lilla28, line 1: is the header needed here?

lilla28 (dismissed):

examples/js-chart-and-grid-messagerouter/js-chart/webpack.config.cjs
kruplm: As we discussed this is a config file, so we don't need to add a license header here.

kruplm:
The issue with Rollup was reported under: https://github.com/rollup/rollup/issues/5701 
I'm waiting until Monday with the merging of this PR.

ZKRobi (approved):

PR #812, opened 2024-10-24
chore: Adding Certified FDC3 2.0 Logo

psmulovics:
Adding Certified FDC3 2.0 Logo


kruplm:
> Where should I put the logo? Should we also write a little blurb about it?

From my side it's ok if you just commit the PNG file, and I update the README. I have some text in mind.

PR #834, opened 2024-11-15
Adaptive cards/notifications

Sanjuice:
This is pull request to expand on the adaptive cards poc with notifications for success, error and information.


ZKRobi:
This is a good start however we need the notifications to appear outside of the web module. Could you extend the PR with this?

examples/adaptive-cards/app.js
```diff
@@ -1,43 +1,102 @@
...
+    return { 
+        fontFamily: "Segoe UI, Helvetica Neue, sans-serif",
+        containerStyles: {
+            default: {
+                "backgroundColor": `${bgColor}`
+            }
+        }}
+}
+
+const renderNotif = (notifType, bgColor ) => {
```
speed-y, line 34: I would prefer to rename the variables to complete names - `renderNotification`, `notificationType`

examples/adaptive-cards/app.js
```diff
@@ -1,43 +1,102 @@
...
+    let notifCard = new AdaptiveCards.AdaptiveCard();
```
speed-y, line 35: `notifCard` -> `notificationCard`

speed-y (changes requested):

Sanjuice:
> This is a good start however we need the notifications to appear outside of the web module. Could you extend the PR with this?

Thanks :) Regarding the notifications appearing using ComposeUI, it would be done in a task by a member from the .NET team so the use case of making the notifications appear outside of the web module would be dealt in that PR. This PR is more regarding the styling of the cards to see how the cards can be styled for different types of notifications.

examples/adaptive-cards/index.html
```diff
@@ -2,10 +2,12 @@
 
 <html lang="en">
         <title>ComposeUI Adapative cards POC App</title>
+        <link rel="stylesheet" type="text/css" href="styles.css">
     </head>
 
     <body>
         <div id="container"></div>
+        <script type="text/javascript" src="https://unpkg.com/markdown-it/dist/markdown-it.min.js"></script>
```
kruplm, line 10: CDNs can time out and be blocked on the firewall.
Could you change the consumption of markdown-it to be through an installed NPM package?

kruplm (changes requested):

kruplm (approved):

lilla28 (approved):
Please do not forget squashing your commits!

PR #867, opened 2024-12-07
Updated docs with the current versions of node and dotnet

misrarah:
Updated docs with the current versions of node 20 and dotnet sdk 8
added a global.json file to enforce the SDK


kruplm:
Hi @misrarah, thank you for the contribution.

The documentation hasn't been updated to contain the current version of .NET, because we haven't upgraded this project to .NET 8 yet.
It's in our pipeline and something we definitely want to do before our next release.

Also, please note that next to your contribution you need to commit a [DCO](https://github.com/morganstanley/ComposeUI/blob/main/dco/dco_template.md)

PR #895, opened 2025-01-03
Realign all Microsoft.Extensions dependencies to v8

bingenito:


Directory.Packages.props
```diff
@@ -22,17 +22,17 @@
...
+    <PackageVersion Include="Microsoft.Extensions.Http" Version="8.0.1" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Options" Version="9.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="8.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="8.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="8.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Options" Version="8.0.2" />
```
psmulovics, line 35: Why not 8.0.1?

psmulovics (approved):

Directory.Packages.props
bingenito: These are all apparently published separately.  For most, 8.0.1 is the latest but for a couple, 8.0.2 is the latest for I assume an issue that was patched.  I grabbed the latest 8 for each dependency (I looked at each one in nuget)

PR #952, opened 2025-02-05
Downgrade FDC3 Examples and Prevent Upgrades to 2.1 Packages

kruplm:
**Changes in this PR**

1. Conforming to FDC3 2.0 Standard
    * Grid example in fdc3-chart-and-grid has been downgraded to 2.0.3
    * The rest of the examples have been upgraded to the latest patch (2.0.3) while making sure that we only allow patch upgrades (~)

2. Preventing unintended upgrades via Dependabot
    * Ignoring Upgrades to 2.1 packages for both NPM and Nuget eco-systems

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,9 @@ updates:
     directory: '/src/fdc3/dotnet/DesktopAgent/src/DesktopAgent'
     schedule:
       interval: 'monthly'
+      ignore:
+        - dependency-name: 'Finos.Fdc3*'
+          versions: [ '>=2.1' ]
     groups:
       infragistics:
         patterns:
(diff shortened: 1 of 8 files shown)


psmulovics (approved):

ZKRobi (approved):

.github/dependabot.yml
```diff
@@ -14,6 +14,9 @@ updates:
     directory: '/src/fdc3/dotnet/DesktopAgent/src/DesktopAgent'
     schedule:
       interval: 'monthly'
+      ignore:
+        - dependency-name: 'Finos.Fdc3*'
+          versions: [ '>=2.1' ]
     groups:
```
phn-ms, line 20: Did I miss something or the FluentAssertions are no longer ignored , like 
```ignore:
      - dependency-name: 'FluentAssertions*'
        versions: [ '>=8.0' ]

phn-ms: My bad. It has never been implemented in the conformance branch.

PR #956, opened 2025-02-20
Handling Popup window's script injection 

lilla28:
Changes:
- First commit : Added `IWindowPolicy` with `DefaultPopupWindowPolicy`  to not inject scripts into popups - opened window via the `window.open` JS call and track the `Window`'s and  `ContentPane`'s child popup windows to enable closing windows if their parent is closed in case of popups.
- Second commit: PLEASE BE AWARE **I am happy to drop this commit** I was just doing some little experiment/research on the `Testably` and `TestableIO` upgrades.
- Third commit: Fixing the examples due to the highcharts upgrade : https://github.com/morganstanley/ComposeUI/commit/06ebda823c883f2758fda574a0df74df28a13cf5

Also please be aware I have an another solution for the script inejction popup windows where I am using the `WebWindowOptions` to create a `FloatingOnly` window instead of creating a `PopupWindow` class. Commit can be found here: [https://github.com/morganstanley/ComposeUI/commit/f41a19950dfa32cd5acc2682467650f0e039eb5f]


psmulovics (approved):

src/shell/dotnet/Shell/App.xaml.cs
```diff
@@ -233,33 +233,44 @@ private void OnAsyncStartupCompleted(StartupEventArgs e)
...
-                    IconUrl = webWindowOptions.IconUrl == null ? null : new Uri(webWindowOptions.IconUrl)
-                }
-            });
-
-            var moduleLoader = _host.Services.GetRequiredService<IModuleLoader>();
-            moduleLoader.StartModule(new StartRequest(moduleId, new List<KeyValuePair<string, string>>
-            {
-                new(WebWindowOptions.ParameterName, JsonSerializer.Serialize(webWindowOptions))
-            }));
+            CreateAndStartNewWebModuleAsync(
```
ZKRobi, line 236: Is this intentionally not awaited?

src/shell/dotnet/Shell/Layout/LayoutManager.cs
```diff
@@ -205,15 +217,25 @@ private async Task LoadModulesAsync()
...
-                        var tcs = new TaskCompletionSource<bool>();
-                        _moduleReadyTasks[moduleId] = tcs;
-
-                        await _moduleLoader.StartModule(new StartRequest(moduleStates[moduleId], new Dictionary<string, string>
-                    {
-                        { SerializationId, moduleId }
-                    }));
-
-                        moduleTasks.Add(tcs.Task);
+                        try
```
ZKRobi, line 220: I'm not convinced this does what you intended it to do - wouldn't this still fail if the individual tasks failed?

src/shell/dotnet/Shell/App.xaml.cs
lilla28: I assume so. this is still just the same functionality with the organization into a new method.

src/shell/dotnet/Shell/Layout/LayoutManager.cs
lilla28: I have reverted back :)

ZKRobi (approved):

PR #962, opened 2025-02-28
Deploy packages permission scope

mimiflynn:


bingenito (approved):
This is experimental and I'm fine with merging. Longer term we need to lower the default permissions.

PR #1001, opened 2025-04-14
Move ADRs to documentation site

mimiflynn:
This migration required a format change to `.mdx` and now includes additional metadata. Happy to discuss.


site/package.json
mimiflynn, line 1: Updating the theme to the version that includes ADR support.

phn-ms (approved):
The paths in the .sln files have been fixed.

PR #1002, opened 2025-04-24
feat(Shell): Add  FDC33 ChannelSelector

kruplm:
This PR introduces a new UI element that allows users to select the FDC3 Channel.
The ChannelSelector is leveraging WPF Material Design PopupBox, and rendered as an overlay thanks to the new [WebView2CompositionControl](https://learn.microsoft.com/en-us/microsoft-edge/webview2/release-notes/?tabs=dotnetcsharp#102957106) which solves the AirSpace Issue

Changes in this PR
* The Embedded WebBrowser is using **WebView2CompositionControl** as a drop-in replacement for the WebView2 Control. As this is using Direct3D under the hood the target framework had to be changed to **net8.0-windows10.0.17763.0**
* We register a communication channel for each instance dynamically so that each web module can be notified by it's channel selector control about the channel change and vice-versa.
* Fdc3ChannelSelectorStartupAction: the control is only added to the DOM if FDC3 is enabled.

### DIFF
--- a/src/shell/dotnet/test/Shell.Tests/Shell.Tests.csproj
+++ b/src/shell/dotnet/test/Shell.Tests/Shell.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>$(TargetFrameworkVersion)-windows</TargetFramework>
+    <TargetFramework>net8.0-windows10.0.17763.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <IsPackable>false</IsPackable>
 	<RootNamespace>MorganStanley.ComposeUI.Shell</RootNamespace>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -29,9 +29,12 @@
     </PropertyGroup>
 	
 	<PropertyGroup>
-		<TargetFrameworkVersion>net8.0</TargetFrameworkVersion>
+		<TargetFrameworkVersion>net8.0-windows10.0.17763.0</TargetFrameworkVersion>
 		<TargetFramework>$(TargetFrameworkVersion)</TargetFramework>
 		<LangVersion>latest</LangVersion>
         <Nullable>enable</Nullable>
 	</PropertyGroup>
+    <PropertyGroup>
+        <WindowsSdkPackageVersion>10.0.17763.53</WindowsSdkPackageVersion> 
+    </PropertyGroup>
 </Project>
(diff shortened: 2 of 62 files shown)


psmulovics:
> We register a communication channel for each instance dynamically so that each web module can be notified by it's channel selector control about the channel change and vice-versa.

Not sure I like this part, let's move this back to draft till we figure this part out

kruplm:
> > We register a communication channel for each instance dynamically so that each web module can be notified by it's channel selector control about the channel change and vice-versa.
> 
> Not sure I like this part, let's move this back to draft till we figure this part out

I have to register a message router topic for each window, as each window has it's own Channel Selector UI element: https://github.com/morganstanley/ComposeUI/pull/1002/files#diff-627415142a469fdf70131cc3a0bf3074fea85e06e376e1b7c625d34f57704697R54

And I need to communicate with these instances. That's what this bullet point means. I can't update this on the main process.

src/shell/dotnet/test/Shell.Tests/Shell.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>$(TargetFrameworkVersion)-windows</TargetFramework>
```
ZKRobi, line 4: If we really need to hardcode this specific version, please keep a central parameter for it and use it in both the shell and the shel.tests project.

ZKRobi (changes requested):
I put a lot of comments around the PR, I'm triing to summarize the main issues here. The comments in the PR will provide more detail or pointers around the issues described below.

There are dependency loops introduced in 2 places:
    - The ModuleLoader and connected infrastructure cannot know about Plugins like FDC3. This is a fundamental architectural choice we made early on and it is not to be changed.
    - The ChannelSelectorClient in the ts part is initialized by the desktop agent but then accesses the desktop agent. This should be revised. What comes to my mind as a solution that sounds "web-like" and keeps the dekstop agent 'web-free' is passing on a callback to handle the channel change messages.
    
Communication protocol and internal logic is mixed with display information. The color of a channel is completely irrelevant for every component except for the actual UI, channel selection should be purely based on the channelId - which might be arbitrarily chosen as the same string representing the color of the channel but it is not the same thing and should not be mixed together.

The lifecycle of the channel selector infrastructure is unclear and/or flawed. Disposable logic is partially implemented (although incorrectly), but never executed. DI registration is done, but is never used. The InstanceCommunicator is implemented in a way that suggests handling multiple topics, but then is instantiated separately.

The MVVM implementation is also messy with responsibilities leaking between the different parts - this is partially tied together with the lifecycle as well.

The design while looking fancy is disrupting the usabilty of modules underneath and is not what we discussed originally. Inside the comments I point out technical reasons as well why I would abandon this design.

There are no tests implemented for the c# parts even though some of it would be very testable.

PR #1009, opened 2025-05-06
Platform Updates for Node and .NET

bingenito:
* Update to Node.JS 22
* Update to .NET 8 including hoist of targetframework to Directory.Build.props and out of each csproj.  Override remains for net8.0-windows


kruplm:
@bingenito we upgraded the tooling as part of the FDC3 2.0 Compliance feature branch. We agreed to do this so that it won't cause discrepancies when we merge the FDC3 2.0 feature branch.

psmulovics:
So let's merge #981

PR #1346, opened 2026-06-09
Migrate to .NET 10 and update GitHub workflows

psmulovics:
This PR upgrades all .csproj files to target .NET 10, updates GitHub workflows to use .NET 10, and reviews solution files for compatibility. Build and test validation is blocked due to a long-running or stuck build process; manual investigation may be required.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        dotnet-version: [ '8.0.x' ]
+        dotnet-version: [ '10.0.x' ]
         node-version: [ '24.x' ]
     steps:
     - name: Checkout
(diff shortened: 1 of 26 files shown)


ZKRobi:
> This PR upgrades all .csproj files to target .NET 10, updates GitHub workflows to use .NET 10, and reviews solution files for compatibility. Build and test validation is blocked due to a long-running or stuck build process; manual investigation may be required.

This is probably stuck because branch protection is set up on the .net 8 matrix combination that does not exist in your PR any more - naturally. This needs to be updated to be able to merge.

ZKRobi:
Either as a follow-up, or in this PR, we will need to remove the version restriction on Microsoft packages

psmulovics:
> This is probably stuck because branch protection is set up on the .net 8 matrix combination that does not exist in your PR any more - naturally. This needs to be updated to be able to merge.

This is now fixed

psmulovics:
> Either as a follow-up, or in this PR, we will need to remove the version restriction on Microsoft packages

You mean in dependabot and such? Yeah.

ZKRobi (approved):

